PR #41, opened 2022-01-20
Tidy up install scripts

marcosci:


ivokwee (approved):

ivokwee (commented):
Not confirmed. Path is supposed to be relative to shiny folder at runtime. Perhaps this is valid for running requirements.R but it would break the app.R.

PR #46, opened 2022-02-15
UNIgene object

layal-khayal:
comment UNIgene because it is removed from the new version of HS an Mm database too


ivokwee (approved):

ivokwee:
OK. Confirmed

ivokwee:
Merged

PR #105, opened 2022-03-31
enable google tags; put ID in Renviron

ivokwee:
Enable Google Tag Manager to allow Google Analytics. GTM is bit more general than just enabling GA. So we use this for adding future tags.


ivokwee:
Should work. Merging in master because we need statistics asap.

PR #114, opened 2022-04-09
change folders for dataview_board

ivokwee:
I think if we decide to put each plot into a separate file we should put them under each board. I just made an example how I think it should be.


ivokwee:
I guess can be closed as we decided for a flat folder structure.

PR #124, opened 2022-04-19
Modals - remove shinyBS

JohnCoene:
This is a cheap trick to remove shinyBS, particularly its `bsModal` function which no longer works (Bs3 vs. Bs5).

@ivokwee could you take a look, I think some do not work, either my port was wrong or they did not work before.

Create a trigger (button) with `modalTrigger`, the second argument is the `id` of the modal it should open, the modal is created with `modalUI`.


ivokwee:
OK. Seems to work a bit but not for all figure/tables. That still needs to be solved, it may be also the old "bug" for the base plots, this was related with the two renderPlots and active binding (pryr) thing. But also some tables don't show. Perhaps create a new issue after this.

PR #128, opened 2022-04-22
Pgx as reactiveValue

ivokwee:
To pass the active pgx object around, currently a handle to a reactive function that is actually created inside the loaiding module, is passed using an "envitonment" list, passed into the modules and env[['load']][['inputData']]. This handle is used inside the modules often referred (and dereferenced) as  ngs <- inputData() or pgx <- inputData, and after that ngs/pgx are used to access the object. I now created a reactiveValues object in the main server part, which is a "list to reactiveValues" from the active pgx objects. Instead of the construct above, the modules can directly access the reactiveValues object referring it as pgx, so without having to do pgx <- inputData(). The major advantage is that the pgx object is also writable, so we can make modifications in the object even in the modules (have to be careful though!). Please have a look in the server.R right after the loading module, and the code changes in board.dataview. The 'env' list object and 'pgx' reactiveValues will remain both existing in the transition, later we can get rid of the 'env' object I think.

### DIFF
--- a/components/app/R/OPTIONS
+++ b/components/app/R/OPTIONS
@@ -17,6 +17,5 @@ TIMEOUT         = 0
 HONCHO_URL      = http://localhost:8000
 
 #BOARDS_ENABLED = load,view,clust,expr,enrich,isect,func,word,drug,sig,scell,cor,bio,cmap,tcga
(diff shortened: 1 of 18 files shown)


ivokwee:
See comment above. I also converted the main module to the new style moduleServer. See again board.dataview.

PR #138, opened 2022-04-23
maximimizing plots in fullscreen zoom

ivokwee:
I made the fullscreen zoom to work for the figures/plots. Can you check?


ivokwee:
Never mind. It's trivial. I am merging it.

PR #140, opened 2022-04-24
Pgx object as  reactivevalues

ivokwee:
Move pgx data object from LoadingBoard to session (server) top level as reactiveValues. Modules can now access directly the PGX object if they wish instead of calling inputData().


ivokwee:
sorry wrong

PR #172, opened 2023-01-06
Plotly styling

z3tt:
Changes to plotly charts across the data view and clustering boards.

Issues that I've encountered or need some expert knowledge and other things that I have noticed are tagged with `TODO` and `NOTE`. The same is reported (with a bit more detail) in the `dev-dataviz` Slack channel.


ESCRI11 (approved):

ESCRI11:
All good, there will be no problems with the PlotModule, it can be merged

PR #174, opened 2023-01-11
board.biomarker -> plotModule refactor

ncullen93:
I split the 4 plots from board.biomarker into separate files and converted them to new plotModule format. All seem to be working for me. They are just base plots.

### DIFF
--- a/components/board.biomarker/R/biomarker_plot_boxplots.R
+++ b/components/board.biomarker/R/biomarker_plot_boxplots.R
@@ -13,22 +13,22 @@
 #'
 #' @export
 biomarker_plot_boxplots_ui <- function(id,
-                                         label='',
-                                         height=c(600, 800)) {
+                                       label = "",
+                                       height = c(600, 800)) {
   ns <- shiny::NS(id)
-  info_text = strwrap("These boxplots shows the expression of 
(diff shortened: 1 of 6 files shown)


ESCRI11 (dismissed):
Fantastic work. Just one small thing left to do. Please add the big `i` button on the top of the board (picture attached for reference)

ESCRI11 (changes requested):
Fantastic work. Just a couple of thing left to do. 

- [x] Please add the big `i` button on the top of the board (picture attached for reference)

image
To do so, please take a look at the `board.dataview` [server](https://github.com/bigomics/omicsplayground/blob/develop/components/board.dataview/R/dataview_server.R#L33-L59) / [ui](https://github.com/bigomics/omicsplayground/blob/develop/components/board.dataview/R/dataview_ui.R#L213-L216)

In this particular board, there is no prior text for this box. Again, following `board.dataview` I would use [this text](https://github.com/bigomics/omicsplayground/blob/develop/components/board.biomarker/R/biomarker_ui.R#L39-L40) 
@ivokwee Maybe you can provide something better?

- [x] Move the` (a)...(b)...` text box to the bottom without any additional description of the board

PR #189, opened 2023-02-12
Fixes before release - xavier

ESCRI11:
- Figure labels are now after the plot title
- Tables have uniform row color (no grey stripes)
- Removed “Data name selected” label at top right of every board
- Removed action button on UMAP figure on load dataset screen
- Moved tableModule to a separate file in order to separate things and work on that
- Fixed scrollX for tables, now the header also moves with the scroller
- Now tableModule has the argument `selector`, placed it accordingly on the tables that have to be selectable


mauromiguelm:
Hi Xavier, really nice improvements! My only comment is that maybe we should check and remove the corresponding css styling code for id ='navheader-current-dataset', so we dont have conflicts in the future with this id.

PR #193, opened 2023-02-20
Develop fix load dataset

ncullen93:
This pr fixes the issue where users clicking "try example dataset" didn't always work. It should always work now because the actions are occurring on the server side via shiny reactive values


mauromiguelm (approved):
I think this is a great solution, I tested the load page buttons and it worked perfectly.

It would be nice to try to understand why `$('#init-example-data').on('click` was not working in some cases and learn from it. Maybe this on the workshop, I would love to have a more in depth look!

PR #204, opened 2023-03-08
Dev settings container

mauromiguelm:
Change listener of settings sidebar from click to mouse over.

- The  bigomics PR is linked to the PR on bigdash with the same name
- Refactor code; remove toggles with more clear declarations
- Remove click listeners to mouseover event, for snappier UX
- keep the same behavior (mobile, timeout, etc..)
- add lock button to give user to lock settings sidebar


mauromiguelm:
> * The navbar should be closed by default on the data loading page but doesn't seem to be. The settings bar is also open by default but I think we agreed that it too should remain closed by default since we need to maximize space in the data loading page for the table + plots
> * The navbar should also be closed by default in the main app but doesn't seem to be. Not sure if we agreed on that.
> * There doesn't seem to be a way to lock the settings bar... the lock isn't showing up

This PR is for right sidebar, I am working on the left sidebar is another branch, as they are independent. I will definitely incorporate these changes in the other PR! For the left sidebar, these are fixed upon running the makefile.

mauromiguelm:
done, with some changes directly added on bigdash

Issue #229, opened 2023-03-14
compute tab not appearing if samples and counts only is uploaded


ivokwee:
compute tab not appearing if samples and counts only is uploaded. should appear at first contrast added. Also suggested contrast name is not updated in the name.

PR #252, opened 2023-03-21
fix: contrasts tab now has auto-suggestion and shows compute tab

ncullen93:
From https://github.com/bigomics/omicsplayground/issues/229


ivokwee:
perfect. seems to work.

ivokwee:
OK

PR #253, opened 2023-03-21
Added cards inside `PlotModule`

ESCRI11:
## Description
Provide the option of using cards inside a `PlotModule` (Figure 1).

## Features
If we place two plots inside the same `PlotModule` I expect them to be simiar in nature, therefore, the information button will display the same no matter which plot is active. The download button will download the active plot and the options button will display the options for the selected plot.

When displaying the modal output, cards are still inside (Figure 2). This could be changed so that the modal shows the selected card only.

## Usage
On the `PlotModuleUI` we have to use the arguments `cards` and `card_names`:

- `card`: Bool value. If `TRUE`, cards will be used.
- `card_names`: Name of the cards (on Figure 1 would be `c("card1", "card2")`).

UI example code Figure1: 

```
PlotModuleUI(
    ns("pltmod"),
    plotlib = "plotly",
    info.text = info_text,
    download.fmt = c("png", "pdf", "csv"),
    title = "Dataset explorer",
    cards = TRUE,
    card_names = c("card1", "card2")
  )
```

On the `PlotModuleServer` we have to create as many servers as cards. In this example, we have two plots, so two servers:

- ID is shared between the UI declaration and all the servers
- We just specify which servers correspond to which card numerically

```
PlotModuleServer(
      "pltmod",
      plotlib = "plotly",
      func = plot.RENDER,
      func2 = modal_plot.RENDER,
      card = 1
  )
PlotModuleServer(
      "pltmod",
      plotlib = "plotly",
      func = plot.RENDER2,
      func2 = modal_plot.RENDER2,
      card = 2
  )
```

## Additional information

I am working on a [developers guide](https://github.com/bigomics/omicsplayground/wiki/Developer-guide:-PlotModule) for the `PlotModule` explaining all the different options it has and how to use them. The provided information in this PR will be included on the wiki.

---

### Figure 1
image

### Figure 2
image


ivokwee:
should we merge now or after Mauro's grid stuff?

ESCRI11:
Yes, the heatmap would be a good candidate to try this out. 

I was planning on waiting for Mauro grid stuff, but we can push this now too.

mauromiguelm:
I like this a lot, going to test it today! There are many interesting applications that comes to mind.

I think we can merge first `Dev grid cards` " #256  into master, and then merge this PR after that.

Or we can merge this PR into `Dev grid cards` " #256, and later to master.

Not sure what is more efficient, but we should do this asap before creating more conflicts.

mauromiguelm:
I couldnt see any tab in loading module, its not yet activated in that plot, right? The PR is just for the backend option to have it I guess, or maybe its implemented and its just me that cannot see?

An interesting option would be to place the tab UI directly in the card header, see this example: https://rstudio.github.io/bslib/articles/cards.html#multiple-tabs

ivokwee:
Best is to be resubmit I guess. File names and locations have changed since the playbase merging.

ESCRI11:
To be tested carefully (also by me)

ESCRI11:
This PR will close #296

mauromiguelm (commented):
I really like the structure behind this, I have some minor comments to try to improve the current approach:

- it would be nice to pass the plotting function "func" as a vector of multiple elements in server, where each one map to card_names in UI. That way we dont have to use `PlotModuleServer` twice.
- Vertical distance in cards is reduced for showing card names, it would be nice to have a more minimal UI.
- Sometimes option changes in Dynamic are not reflected in Static, for example the row clustering shows genes only in dynamic, but it varies in the static, with static having multiple labels. I think plots should show consistant information, and this might be a legacy issue unrelated to your PR. I guess an easy solution we could hide static row dendogram in case the clustering is the same.
- Plots do not get to the bottom of card, this might have something to do with the fill behavior of the parent div, we can discuss a bit more if you wish, I think a solution is easy.
image

ESCRI11:
Ok point 2 can be achieved easily, I have a working prototype (ugly but functional)

ivokwee:
> Ok point 2 can be achieved easily, I have a working prototype (ugly but functional)

Can you show? My idea was to use bslib::navs_hidden (see  https://rstudio.github.io/bslib/reference/navs.html) and add somehow an actionButton in the drowdownmenu (or header?) that does something like   `updateTabPanel("card_selector", selected = "card1")`

ESCRI11:
@ivokwee Check the latest commit [8d1b84f](https://github.com/bigomics/omicsplayground/pull/253/commits/8d1b84fc76b4f03fbff3272b44b5faaef2bccc22). Basically I grab the "primitives": `navList` and `content` and place the on the header and body sepparately. I just have to make them inline and not affect the buttons.

ESCRI11:
Yep exactly, I am trying to figure out the best solution

ESCRI11:
@ivokwee @mauromiguelm Please feel free to try the latest developments. Only thing missing is points 3 and 4, which should be easy.

image

ESCRI11:
Only missing fixing the heatmap options so that in both plots are relevant.

mauromiguelm:
I really like this PR, it looks beautiful Xavier! I tested some of the option and it works quite nice, is there anything broken still?

I did two things (not yet pushed):

- I reduced the font size of the tabs, do you like it?

image

- create a small look in the server so we dont call plotmoduleserver twice.. I think its a big more easy to understand, especially if we decide to have multiple plots in the future!

Let me know if I should push these two updates!

ESCRI11:
Great! Good eye with the smaller font size. Please push the changes. I think we are ready for merge

mauromiguelm:
Done, I checked again and all looks good!

Problems related to the heatmaps:

- absolute versus relative coloring between the two heatmaps are quite off, we should match color and outcomes as much as possible between the two heatmaps.. 
- show legend is not working when we select split by phenotypes

This is kind of unrelated to this PR, we could merge and keep working these issues on #296, as we have other priorities

ESCRI11:
Agreed. Let's finish this PR and solve heatmap problems as different issues

PR #256, opened 2023-03-22
Dev grid cards

mauromiguelm:
**Achievements:**

- implemented cards on plotmodule
- styling cards that were overwritting bslib defaults
-  added legends to cards, default to info-text. All boards will need to be revisited to edit the legends and remove bottom of page legends
- plot legends now have same height across page
- option to hide and show legend added across all boards
- styles match our style guidelines

**Ongoing development:**

- fix Gene information > Gene overview coliding with plots
- allow plot fill for cards > 400px.. this is difficult, in talks with Carson, should be patched soon


mauromiguelm:
cards are ready for merge, fixed were pushed.

ivokwee:
I think the name 'selector_default' should be more specific like 'caption_switch'. Also the jquery selector on 'form-check-input' is too general and may clash if you use the widget for something else. We need to add a class like 'caption-onoff' or 'caption-switch' just to avoid future clashes.

mauromiguelm:
> I think the name 'selector_default' should be more specific like 'caption_switch'. Also the jquery selector on 'form-check-input' is too general and may clash if you use the widget for something else. We need to add a class like 'caption-onoff' or 'caption-switch' just to avoid future clashes.

I will give the option to define classes as args, so we can re-use it again in the app in the future for different purposes. I gave a general name to allow reusability.. hope thats ok!

components/app/R/www/temp.js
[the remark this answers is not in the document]
ivokwee: the PR had this code where addClass-d-none is added when shown, d-none means 'display: none' so hiding.

image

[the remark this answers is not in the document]
ivokwee: this seems ok 
image

components/base/R/ui-PlotModule.R
ivokwee, line 1: The Spinner should be put back. Or any reason?

components/base/R/ui-code.R
ivokwee, line 1: again. selector_default is not a good name, is you use this just for the caption then caption_switch or if you want to retain this as general widget select_switch but then add extra class like 'caption-switch' when you use it.

mauromiguelm: will do!

components/base/R/ui-PlotModule.R
mauromiguelm: nop, I think I copied the test code or got lost somewhere.. will be back!

ivokwee:
Better put selector_default in boardHeader().  I tried  no problems with JS. You even don't need an inputID. You might need an extra div or class for selecting in CSS.
```
boardHeader <- function(title, info_link) {
    fillRow(
        flex=c(NA,1,NA),
        ##h2(input$nav),
        shiny::div(
            id = "navheader-current-section",
            HTML(paste0(title," &nbsp;")),
            withTooltip(
                shiny::actionLink(
                    inputId = info_link,
                    label="",
                    icon = shiny::icon("info-circle"),
                    style = "color: #ccc;"
                ),
                "Show information and tutorial about this board"
            )
        ),
        br(),
        selector_default("", label = "Show captions")
    )
}
```

mauromiguelm:
Comments addressing and fixing the last round of discussion:

- classes on selector_input are BS5 classes that were wrapped to work in shiny (exemple is form-check-input), so I cannot remove them. I added the option to add classes, so the component can be re-used.
- selector_input moved to navheader
- spinner completely disrupts height definition from `htmltools::bindFillRole`, we wil need to find another solution. Another solution is actually a good thing as spinners are running in background, as you mentioned Ivo!
- selector_input now shows on all board BUT loading board, as loading board does not have many plots and it looks ugly

PR #278, opened 2023-04-05
Refactoring captions

mauromiguelm:
- early PR to merge changes on plot module and boards
- New format of info.text and captions
- 2x tabs in Dataview converted to new format for testing, other plots will have no caption **(!!!)**
- Bring title, info.text, caption, height and width of each plot to module UI.


ivokwee:
Merged. I understand some captions are still to be done/updated. Let's do that ASAP.

PR #294, opened 2023-04-13
feat: double plot download resolution

### LINKED ISSUE #291: plotly exported PNG are low resolution
Plotly exported PNG are low resolution. For example the cluster sample > phenotypes is 640x960. Would be good to have higher resolution. Probably twice. Perhaps need to set this at exporting with Kaleido.

ESCRI11:
This closes #291 

## Description

I updated the pixel size operations to be doubled. That way a plot that by default was downloaded with a resolution of 640x640 now is 1280x1280.


ivokwee:
OK for now. We actually have the "res" parameter in the PlotModule that increases resolution like 80dpi or 150dpi. Seems it was not used for plotly export. For other formats it was applied. Let's keep this solution for now.

PR #305, opened 2023-04-13
Fix #272 pca/tsne 3D plot not working

### LINKED ISSUE #272: pca/tsne 3D plot not working
The pca/tsne 3D plot are not working anymore.

ESCRI11:
This closes #272 

## Description
The file `clustering_server.R` was trying to access an input value declared on `clustering_plot_PCAplot.R`. Updated the input selector to fetch the correct one.

Also, there was a couple of problems handling the 3D data, fixed those.

- Pass correct columns. [Code](https://github.com/bigomics/omicsplayground/commit/1db92a178ddd7c144d3b2dce71e181a72b6de779#diff-dde611e40d231c9ab781e9ee7826eb9a83fa68c08229d9f59aa8a898528f3fc5R76-R80)
- Assign proper column names. [Code](https://github.com/bigomics/omicsplayground/commit/1db92a178ddd7c144d3b2dce71e181a72b6de779#diff-57d6a9e3f42e5bc202e40194cf58eaa81fc6fabd26d8137381deb233d36d65e8R51-R55)


ivokwee:
oh.... passing an input from a child to the parent with explicit namespace.... I let Nick decide how to best elegantly solve this... @ncullen93 can you look and decide?

ESCRI11:
@ivokwee The reason this is happening is because the Phenotypes plot is affected by this input value. One option would be to place the option on the right side settings.

ivokwee:
OK. Let's accept for now. Good hack but perhaps bit ugly. Communication between modules was one topic at the Appsilon conference...

PR #323, opened 2023-04-19
Dev margin plots

mauromiguelm:
- remove margin left and right of plots
- increase information density, a common request for v3 users


components/ui/ui-PlotModule.R
ivokwee, line 1: why not remove also top and bottom? I would put it in the plotly_default() nor in the plotmodule and still let the dev still be able to override the margin.

mauromiguelm:
I agree, here is a workaround:

**problem:** Few plots <10% have pllotly_default, so we will need to go plot by plot 

**solution:** I added an argument to plotModule where the user can opt to remove margins. In the (rare?) case we see a plot that requires margin, we set that to false and add margins manually.

**necessary additional steps:** if this solution is approved, there are few plots that we actually need to add margin. I will go one by one and fix that.

image

ivokwee:
Solution OK. Please fix the plots that require margins.

PR #327, opened 2023-04-19
Follow up #324 + fix non issued bug

ESCRI11:
Following from #324 

## Description
@mauromiguelm  Good catch. I have worked the changes inside `getGeneCorr`, which removes the `pcor_ntop` input for `correlation_plot_table_corr`.

Similarly I removed unnecessary subsetting on `correlation_plot_scattercorr`.

Also, I have noticed the Gene selector was not acting properly on "Correlation scatter plots" from the same board (Fig 1), fixed it to act as it should (Fig2)

### Fig 1
image

### Fig 2
image

### DIFF
--- a/components/board.correlation/R/correlation_plot_scattercorr.R
+++ b/components/board.correlation/R/correlation_plot_scattercorr.R
@@ -82,7 +82,6 @@ correlation_plot_scattercorr_server <- function(id,
     })
 
     cor_scatter.PLOTFUN <- function() {
-      browser()
       dt <- cor_scatter.DATA()
       rho <- dt[[1]]
       this.gene <- dt[[2]]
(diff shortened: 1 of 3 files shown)


ivokwee:
Sorry guys. I did a big refactoring yesterday of this correlation panel. It was a bit messy and indeed the use of ntop was wrong. It is for computing the partial correlation and should not be used for truncating the table R. The table is now the main driver for selection and ordering. I also split up table and plot code, both barplot and scatterplot follow the selection and ordering of the table. I think these changes are obsolete now. Please check the current version already in the master. 

Closing this issue.

ivokwee:
@ESCRI11  I didn't merge your commits because the panel changed a lot. Can you check if your changes are still relevant and if so include those to the new version already in master?

ivokwee:
@ESCRI11  I didn't merge your commits because the panel changed a lot. Can you check if your changes are still relevant and if so include those to the new version already in master?

ESCRI11:
Just checked. With the changes you made on master this PR is irrelevant. Closing it.

PR #329, opened 2023-04-21
Improving `Geneset enrichment` board

mauromiguelm:
**Problems:**
1. Plots were blinking when changing contrasts
2. UX was strange when scrolling gene set table; user had no ability to select multiple gene sets and compare them

**Solutions:**
1. Reactives are not invalidated anymore when changing contrasts
2. User can select one or multiple gene sets


ivokwee:
This needs more thinking. Multiple selection works well for the first tab 'Top enriched' but does not work for the 'Plots' and 'Compare' tab. I think multiple selection is more confusing than useful.

Issue #289, opened 2023-04-13
Some plots cannot be downloaded PNG/PDF


ivokwee:
These plots cannot be downloaded as PNG/PDF:

Cluster features > Gene > Gene Signatures  (base plot)
Cluster features > Geneset > Geneset Signatures  (base plot)
Correlation analyis > Graph > Partial correlation network (visnetwork)
Correlation analyis > Graph > Partial correlation network (visnetwork)
Geneset enrichment > Top enriched > Top enriched gene sets (base)
Geneset enrichment > Plots > Volcano plot (base)
Geneset enrichment > Plots > Enrichment vs. expression (base)
Pathway analysis > Pathway Analysis > GO graph
Word cloud > Word cloud > Enrichment plots (base)
Word cloud > Word cloud > Word Cloud (base)
Drug Connectivity   > Drug enrichment > Activation matrix (base)
Compare signatures   > Pairwise scatter > Venn diagram
Similar experiments   > FC Heatmap > Connectivity heatmap (base?)
Similar experiments   > Meta-graph > Leading-edge graph (visnetwork)
Similar experiments   > Meta-graph > Enrichment graph (visnetwork)
Single Cell Board > Cell type > Cell type profiling (ggplot/grid)
Single Cell Board > Markers > Expression of marker genes (ggplot/grid)
Single Cell Board > Markers > Cytometry  plot (ggplot/grid)
Compare datasets (beta) > Compare expression > Dataset1 (base)
Compare datasets (beta) > Compare expression > Dataset2 (base)
Compare datasets (beta) > Gene Correlation > Expression (base)
Compare datasets (beta) > Gene Correlation > Gene correlation (base)
TCGA (beta) > TCGA survival > TCGA survival analysis (base)
Clustering > WGCNA (beta) > WGCNA > Gene dendrogram and gene modules (base)
Clustering > WGCNA (beta) > WGCNA > Scale independence and mean connectivity (base)
Clustering > WGCNA (beta) > WGCNA > TOM heatmap (base)
Clustering > WGCNA (beta) > WGCNA > Gene clustering (base)
Clustering > WGCNA (beta) > WGCNA > Module graph (base)
Clustering > WGCNA (beta) > Modules > Module-Trait relationships (base)
Clustering > WGCNA (beta) > Modules > Correlation network (base)
Clustering > WGCNA (beta) > Modules > Correlation network2 (base)
Clustering > WGCNA (beta) > Eigengenes > Eigengene clustering (base)
[...]

PR #330, opened 2023-04-21
Fix plot download

mauromiguelm:
See last comments on #289.

We have three plots not working, to avoid the risk of breaking the others (no comprehensive tests), I could disable download for the ones that are not working.


mauromiguelm:
Disabled download for the three plots that are not working:

**PDF disabled**
[png=OK,pdf=failed] Single Cell Board > Markers > Expression of marker genes (ggplot/grid) investigate, pdf indeed not working

**entire download disabled**
[failed] Single Cell Board > Markers > Cytometry plot (ggplot/grid) investigate, confirmed not working

**entire download disabled**
[failed] Clustering > WGCNA (beta) > Modules > Module-Trait relationships (base) investigate, confirmed not working

PR #338, opened 2023-04-25
Dev wgcna

ivokwee:
briefly the pgx.getFamilies had a bug. It was recreating the index list for all genesets (takes 5 minutes...). Also because GSETS or iGSETS was not in playdata yet. I have asked an PR for playdata to include GSETS and iGESTS (and more global variables). The lib.dir is not needed in pgx.getFamilies. Actually, the pgx.getFamilies can be retired because the pgx object after pgx.initialize has a families field/slot. This was done becuase families are often called in filters. There are some merge conflicts but I think it should work.


ivokwee:
This PR should be done together with the PR in playdata and playbase.

PR #339, opened 2023-04-25
#335: from tipify to withToolTip

### LINKED ISSUE #335: tooltips in batch correction tab needs to be updated with withTooltip
Tooltips in "batch correction" tab are using tipify2/tipify and show old style tooltips. Elsewhere we use withTooltip (in .../ui/ui-tooltip.R). Please replace here and elsewhere in other modules.

ESCRI11:
This closes #335 

## Description
Replaced all `tipify`/`tipify2` with `withToolTips`

### DIFF
--- a/components/board.clustering/R/clustering_ui.R
+++ b/components/board.clustering/R/clustering_ui.R
@@ -30,9 +30,10 @@ ClusteringInputs <- function(id) {
     shiny::conditionalPanel(
       "input.hm_features == '<contrast>'",
       ns = ns,
-      tipifyR(
+      withTooltip(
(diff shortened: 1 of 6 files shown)


ivokwee:
Xavier. Do you perhaps know, or can you create, a server-side function that adds a tooltip to an element, like `addTooltip('my_input','Hello world!')`

ESCRI11:
I am not sure I follow what you have in mind. Can you please provide some more context on what should be the output of such function?

ivokwee:
OK. Thanks. BTW withTooltips is not actually using options. It was just kept for compatibility/laziness.... see ui/ui-tooltip.R.

PR #345, opened 2023-04-27
flush stderr/stdout from background processx to avoid blocking

ivokwee:
Sometimes the background process seems to halt or blocked. I suspect blocking of stderr/stdout is happening as described here https://processx.r-lib.org/

> Note, that the connections have a buffer, which can fill up, if R does not read out the output, and then the process will stop, until R reads the connection and the buffer is freed.
> Always make sure that you read out the standard output and/or error of the pipes, otherwise the background process will stop running!

So during process checking we need to flush the stderr/stdout buffer to avoid any fill up. I changed the code.

### DIFF
--- a/components/modules/ComputePgxModule.R
+++ b/components/modules/ComputePgxModule.R
@@ -392,40 +392,35 @@ ComputePgxServer <- function(
                     selected = 'load-tab'
                 )
 
-                dbg("[compute PGX process] : starting process")
                 process_counter(process_counter() + 1)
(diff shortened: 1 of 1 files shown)


mauromiguelm (approved):
- tested on wsl, did not block the app as before.
- added a small comment about computation of multiple datasets
- recommend storing the process directly on lists, and not on intermediate variable, as I tried this solution before and it did not work for multiple processess.

PR #357, opened 2023-05-07
Fix #314 Watermarking does not work anymore

### LINKED ISSUE #314: Watermarking does not work anymore
Please enable global watermarking of PNG/PDF images

ESCRI11:
This closes #314 

## Description
Removed the hardcoded watermarks so that the global option works.

Make sure to have installed the Linux `convert` and `pdftk` packages, otherwise the watermarking will just not work.


ivokwee:
Works: Left PNG, right PDF.
image

PR #370, opened 2023-05-10
Insert page UI after dataset load rather than at app start

ncullen93:
This PR makes it so only the welcome and loading pages are loaded when users enter the app. Only once the user loads a dataset do the dataview et al pages get inserted into the UI. This means that actually entering the app is much faster, although the time to wait on the modal screen after loading a dataset is slower. This PR addresses the idea that waiting >10 seconds just to see the first render of the app is not acceptable. This also still allows us to work on reducing the rendering time of the pages.

An extremer version of this idea would be to only insert a page's UI when that page is visited. That works really well with the page servers, in my experience.


mauromiguelm (changes requested):
- the metric "time-to-first-page" is greatly improved, on that sense the PR is passed! I agree there's no simpler solution other than rendering the ui piece at a later stage.
- I noticed that parts of the the `beta features` are not working, my guess is that the switch to the server changed some of the javascript selectors and how we insert new tabs. Could you confirm that this is happening to you?
- If yes, we need to be careful when `switching between datasets` with higher or lower elements in the PGX, as the tabs need to be updated accordingly

mauromiguelm:
> @mauromiguelm can you give me specific examples of which beta tabs aren't working ? It looks OK to me for the most part but maybe I'm missing something.

when beta features were enabled, the beta `wgcna board` was not showing in settings, even if dataset had wgcna computed.

PR #387, opened 2023-05-15
add ability to preview uploaded data

ncullen93:
This PR adds the ability to preview uploaded data in a table via modal dialog. If the user discards the data, it gets NULL'd. 

Currently this only allows preview of counts.csv -- we need to think about how to do this for samples.csv and potentially contrasts.csv. I think we can either try to show them all in one modal dialog and have tabs to switch between datasets, or try to show multiple modals in a row. This requires a bit more thinking, but the general idea of how we do uploaded data preview can still be discussed in the meantime.

### DIFF
--- a/components/board.upload/R/upload_table_preview.R
+++ b/components/board.upload/R/upload_table_preview.R
@@ -49,28 +49,37 @@ upload_table_preview_server <- function(id, uploaded) {
                     tabs,
                     list(tabPanel(
                         'Counts',
-                        div(
-                            shiny::actionButton(
-                                session$ns('convert_coltype'),
-                                label = 'Convert column type',
-                                style = 'display: inline-block',
-                                class = 'btn-secondary btn-sm'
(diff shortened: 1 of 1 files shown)


mauromiguelm (commented):
1. In this case it would be nice to think how this will be used in the future and prepare for that. Some of the functionalities that comes to mind that might benefit upload:

- Highlight errors in csv files and show directly in table. Ideally, we give directions on how to fix it based on the identified problem.
- Allow users to select if a column with metadata is numeric or categorical (factor, char)
- Allow users to add a better name for contrasts. We can still use act_no_act, but the user can add a "publication" name for that, for example Activated vs. Not activated, that will be shown in plots.

2. I would embed the table in the page, not as a pop-up, and have one tab per csv. Maybe the user should go through each tab, maybe not. But if we pick up errors in a certain csv, the check should be mandatory.

Issue #380, opened 2023-05-12
add to user profile -- (1) hide/show bs_alerts and (2) hide/show captions


ncullen93:
There should be an option in user profile to show or hide the bs_alert for each board / tabset, and also to show or hide plot captions. These would apply app-wide. However, I still like the fact that hide/show captions are on each page so that you can just quickly flick it on to see the captions and then close it. So we should consider whether to keep that going forward.

PR #388, opened 2023-05-15
add tab-info-alert toggle to profile page

ncullen93:
This PR provides an option on the user profile/settings page to enable or disable the blue bs_alerts on each page that provide info about the page or tab. These are something the user probably shouldn't see more than once (like when they first use the app) because they are text-heavy (i.e., likely skipped) and take up a lot of vertical space. 

Right now this is only implemented for the very first bs_alert on the dataview page, so that the implementation can be discussed.

Ref issue:
https://github.com/bigomics/omicsplayground/issues/380

### DIFF
--- a/components/app/R/server.R
+++ b/components/app/R/server.R
@@ -200,7 +200,7 @@ app_server <- function(input, output, session) {
 
           if(ENABLED['dataview'])  {
             info("[server.R] calling module dataview")
-            DataViewBoard("dataview", pgx=PGX)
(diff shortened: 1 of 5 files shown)


ivokwee:
Agree. the show captions, toggle alert, and show beta features could be in one "preferences" or "settings" (but we used this name already for the right sidebar). I prefer probably a new menu item in the drop down, rather than inside 'user settings'

PR #389, opened 2023-05-15
Improvements in Enrichment Geneset Board

mauromiguelm:
**Overview:**

The geneset enrichment board is one of the most used boards in the App. That means we need good UI/UX, but also clear descriptions on what each tab within the board does.

**Problems:**
- Severe flickering caused by DT table invalidating certain reactives;
- Some tabs had 50% of the plots updated when clicking in a table, that's a very weird behavior. Ideally, all plots in a tab should be updated, otherwise it feels like something is broken;
- Certain plots had a geneset filter that was defined by the scroll of the geneset table. If I scroll down in the table, certain plots would update based on scroll. That's a very strange behavior that should be removed.
- Certain tabs and plot titles were not clear. 

**Overview:**
- All plots in a tab update when we select a geneset or a gene;
- Flickering problem solved;
- Geneset defined by table scroll behavior removed;
- Title of tabs updated to better reflect plot content; had to move a few plots around to accomplish that;
- Segment tabs by content;
- Added a statistics tab with summary of contrasts and methods, implemented in Tabs within Cards.

**Possible future improvements:**
- Convert single volcano plots to plotly;
- Revert ranking order of GSEA plot to match volcano results (right side positive, left negative enrichments)
- Add volcano plots by contrast in tab `Compare contrasts`
- `Compare statistics` tab: Filter plotted methods by using the selected methods in settings options


mauromiguelm:
> Looks good, but I get actually a short flicker of "Please select a geneset" on the first plot (Genes colored by geneset) and then it shows the figure. If you can fix that, then I think it would be perfect.

definitely! I will wait for Ivo's feedback and make the adjustment. Thanks Nick!

PR #438, opened 2023-06-04
Share pgx files between users

ncullen93:
This PR allows users to share pgx files with other users. It also splits the "share public" and "share with user" functionalities. Users are required to accept all incoming share requests before the datasets are made available to them. Users receive an email that someone has shared a dataset with them (FYI this requires a gmail SMTP credentials file that is not included in the PR). 

NOTE: to test this locally, you can add your email to the CREDENTIALS file in the app directory, but it cannot be a personal email.


mauromiguelm:
Great work nick!

PoC approved, some comments:
1. File name should not end with .com, as windows things its a MS-DOS application.
2. When I compute new PGX, datasets table is not being updated anymore, I need to reload application to see the new dataset.

Cheers,
M

ivokwee:
I am merging this but @ncullen93 please check  getRequestedSharedPGXINFO()  it got a merge conflict and looks strange. Is it used? Why is return a list of files instead of dataframe?

PR #446, opened 2023-06-06
Fix #444 Warning: 'card_body_fill' is deprecated. Use 'card_body' instead

### LINKED ISSUE #444: Warning: 'card_body_fill' is deprecated. Use 'card_body' instead.
Seemingly we need to update the card_body function call.

ESCRI11:
This closes #444 

Trivial fix. @ivokwee Just make sure that Docker `bslib` version is up to date so we don't crash it.

### DIFF
--- a/components/ui/ui-PlotModule.R
+++ b/components/ui/ui-PlotModule.R
@@ -175,7 +175,7 @@ PlotModuleUI <- function(id,
     zoom.button <- modalTrigger(ns("zoombutton"),
       ns("plotPopup"),
       ##      icon("window-maximize"),
-      icon("up-right-and-down-left-from-center"),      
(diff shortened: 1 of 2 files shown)


ESCRI11:
Regarding the elongated plots, it is solved on latest bslib master branch, which given we were getting the warning, it is what is being used on the deploy. This latest release also deprecates two additional functions. Will do the suggested cleaning

ESCRI11:
> > Regarding the elongated plots, it is solved on latest bslib master branch, which given we were getting the warning, it is what is being used on the deploy. This latest release also deprecates two additional functions. Will do the suggested cleaning
> 
> Ah ok so I have to install bslib through github?

Yes

PR #455, opened 2023-06-11
Fix #422: Gene UMAP not displaying mouse gene acronyms in V3.11 and V3 (3.0.RC6.230.426)

### LINKED ISSUE #422: Gene UMAP not displaying mouse gene acronyms in V3.11 and V3 (3.0.RC6.230.426)
Under "Clustering>Features", when using a mouse genome, the gene UMAP does not correctly display mouse gene acronyms. This is actually a legacy issue dating back to V2, as only the capitalised gene names (identical to corresponding human acronyms) are shown:

image

ESCRI11:
This closes #422 

## Description
There was a mismatch due to different capitalization on mouse genes. I have used `str_to_upper` to solve it. I first get the indexes to use the correctly capitalized strings downstream. I am not 100% sure this approach is OK. We can discuss it before merging.

### DIFF
--- a/components/board.featuremap/R/featuremap_plot_table_gene_map.R
+++ b/components/board.featuremap/R/featuremap_plot_table_gene_map.R
@@ -84,7 +84,6 @@ featuremap_plot_gene_map_server <- function(id,
     })
 
     plot_data <- shiny::reactive({
-
       ## pos <- pgx$cluster.genes$pos[['umap2d']]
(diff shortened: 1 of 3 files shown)


ivokwee:
Partly the problem is that FAMILIES is now in playdata and defined as human (fully capitalized) gene symbols. I think previously in v2, FAMILIES was defined for mouse seperately.

PR #456, opened 2023-06-12
Input files cleaning and warnings moved to playbase

mauromiguelm:
**WARNING:** This PR is coupled to the OPG PR of same name. https://github.com/bigomics/playbase/pull/17

- Remove input files cleaning and warnings from the UI and place them into `playbase::playbase::pgx.checkPGX()`.
- If errors are found in the input files, `playbase::playbase::pgx.checkPGX()` will return an error message.
- The return value of playbase::playbase::pgx.checkPGX() is the corrected dataframe (similar to what OPG does currently), with the parameter `PASS` indicating whether errors were found (TRUE means no errors were found), and the `error log` containing a list of all the errors found. Each element in the list represents a different error type.
- The possible error messages are stored in `playbase::PGX_CHECKS`.
- This modification will facilitate the expansion of both input file checks and enable Nick to incorporate these error messages into the dataset preview UI.

### DIFF
--- a/components/board.upload/R/upload_server.R
+++ b/components/board.upload/R/upload_server.R
@@ -267,7 +267,6 @@ UploadBoard <- function(id,
             IS_EXPRESSION <- grepl("expression", fn1, ignore.case = TRUE)
             IS_SAMPLE <- grepl("sample", fn1, ignore.case = TRUE)
             IS_CONTRAST <- grepl("contrast", fn1, ignore.case = TRUE)
-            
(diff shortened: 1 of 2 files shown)


mauromiguelm:
> Looks great. This is exactly what we need. There are merge conflicts because the ComputePgxModule.R file has been moved and renamed but I can take care of that whenever this PR is ready to be merged.

very good points, I will fix it asap

mauromiguelm:
I did not do on purpose! I thought it was merged and deleted from my github desktop...

PR #476, opened 2023-06-14
Define root independently of /omicsplayground

mauromiguelm:
**Description:**
- previously we had to name the folder omicsplayground in order for `get_opg_root` to define the root.
- This is prone to errors, as one could have omicsplayground_repo/omicsplayground. 
- Similarly, a git worktree with root name as branch, the `get_opg_root` would not work and crash.
- Problem was solved for both `Makefile` and `shiny::runApp()` by directing root to the third immediate parent.
- We can now use worktrees efficiently.

**Future:**
- Traditionally root should be where the .Rproj file is, so we could implement a more clean solution that works for both  `Makefile` and `shiny::runApp()`.

### DIFF
--- a/components/app/R/global.R
+++ b/components/app/R/global.R
@@ -46,8 +46,11 @@ TABLE_HEIGHT_MODAL <<- "75vh"
 reticulate::use_miniconda('r-reticulate')
 
 get_opg_root <- function() {
-    pwd <- strsplit(getwd(),split='/')[[1]]
-    paste(pwd[1:max(grep("omicsplayground",pwd))],collapse='/')
+  pwd <- getwd() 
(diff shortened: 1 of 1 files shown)


ivokwee:
Not sure if this is "better". The idea was that you could safely call get_opg_root from any folder below the root. Now it assumes the current active R folder is always .../components/app/R. That is true for perhaps the running app in the docker but not when running some parts of code interactively, for example when debuggin. Actually the max() operator, in `max(grep("omicsplayground",pwd))]`, should take the lowest folder as path. It should take the `.../omicsplayground_repo/omicsplayground` case correctly. If you do not have problems, please "do not fix things that are not broken". Well... sometimes  it's good to fix non-broken things for clarity... only sometimes

PR #478, opened 2023-06-14
Fix #423 compare datasets gene labels in mouse

### LINKED ISSUE #423: Cannot compare Human and mouse datasets under "signatures in V3.11 and V3 (3.0.RC6.230.426)
Under "Signatures->compare datasets (beta)", if I compare a human and mouse genomes, I get no gene names showing up except the ones consisting of a single capital letter. This is an issues carried on from V2, related to the capitalisation of Human but not mouse HUGO acronyms:

image

mauromiguelm:
this fixes #423 

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 libx/
-
+gmail_creds
 
 # R data
 ######################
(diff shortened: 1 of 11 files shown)


mauromiguelm:
> Looks good - just double checking that line 59 of the compare_plot_..R file which has the `browser()` call is necessary or should be removed?

done

PR #504, opened 2023-06-21
Fix frozen welcome page

### LINKED ISSUE #503: Calculation of dataset info (sigdb, allFC, info) blocks server on the welcome page
platform is freezing on welcome page when `datasets-sigdb.h5`, `datasets-allFC.csv`, `datasets-info.csv` are not available.

https://github.com/bigomics/omicsplayground/assets/28757711/fe5bdb2e-c2aa-4007-bb7c-7e5a23a926db

mauromiguelm:
This fixes #503 

**Path to solution:**
-  I tried MANY things with the `getPGXINFO` and `getPGXDIR` in the loading board. Nothing works as these functions are triggered before any button in the welcome page can be clicked. That means we will have to wait 5-60 seconds to see a response from server when clicking any button in welcome page.
- solution was to move `getPGXINFO` and `getPGXDIR` to server, so it gives more flexibility when to call these functions.
- I decided to create an observer that checks if user is logged in, and if yes, it will release a modal in the welcome page. This was not possilble before, when these functions were in loading board.
- Note that this issue only happens in special cases where the `datasets-sigdb.h5`, `datasets-allFC.csv`, `datasets-info.csv` need to be updated.

**Possible future developments:**
- separate UMAP and tSNE calculation from these functions, so that UMAP, tSNE are only triggered when really needed. Which could be loading board table clicked (?). This should make welcome page be released quicker, postponing the computation overhead.
- Run the  `getPGXINFO` and `getPGXDIR` as background computation, as they can take some time.
- Bring the  `getPGXINFO` and `getPGXDIR' even earlier, when user first authenticate. This is the earlist I could get without having to do a lot more dangerous things.
- Find a better place to call `getPGXINFO` and `getPGXDIR`, maybe when we click any button in welcome page? This will need a lot of evaluation, as there's a lot of reactivity behind.

**tests:**
- logged, not logged
- starting OPG with and without `datasets-sigdb.h5`, `datasets-allFC.csv`, `datasets-info.csv`
- adding a pgx in data/

**Please help with tests, as this is a critical module in OPG!**

### DIFF
--- a/components/app/R/server.R
+++ b/components/app/R/server.R
@@ -186,12 +186,10 @@ app_server <- function(input, output, session) {
         }
         
         info("[SERVER:getPGXINFO] updating datasets-info.csv")
-        pgx.showCartoonModal()
-        try(getPGXINFO())
-        shiny::removeModal()
-        
+        pgx.showSmallModal()
+        getPGXINFO()
+        shiny::removeModal(session)
     })
-
 
     ## Global reactive values for app-wide triggering
     r_global <- reactiveValues(
--- a/components/app/R/server.R
+++ b/components/app/R/server.R
@@ -176,6 +176,22 @@ app_server <- function(input, output, session) {
       info
     })
 
+    shiny::observeEvent(auth$logged(), {
+        ## trigger on login
+        logged <- auth$logged()
+
+        if (!logged) {
+          warning("[SERVER:getPGXINFO] user not logged in!")
+          return(NULL)
+        }
+        
+        info("[SERVER:getPGXINFO] updating datasets-info.csv")
+        pgx.showCartoonModal()
+        getPGXINFO()
+        shiny::removeModal()
+        
+    })
+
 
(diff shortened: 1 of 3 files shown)


mauromiguelm:
Hi Nick,

1. **loading page:** Changing the reactivity to the table loading will require some work, as we need these functions called in order to load example data button, as we dont know if example data is available in datasets or not. I tried this but got many problems given reactivity complexity.
2. **data shared** I will move those functions to the server as well. I assume equal unnacceptable long time for the shared datasets, so it makes sense to add the modal in those cases.

mauromiguelm:
because its shiny::eventReactive(), I could add modal directly in the loading server. Could you test with the shared options?

mauromiguelm:
it works with pgx.showSmallModal() but not with pgx.showCartoonModal() for some reason. I tried many things, but SmallModal will do the job. I will double check and bring back to server.

M

PR #505, opened 2023-06-21
Fix frozen welcome page v2 (simplified)

mauromiguelm:
This fixes https://github.com/bigomics/omicsplayground/issues/503

Path to solution:

- I tried MANY things with the getPGXINFO and getPGXDIR in the loading board w.r.t. reactivity. Nothing works to delay these functions are triggered before any button in the welcome page can be clicked. And adding response to buttons is not a good way to solve this.
- simplest solution was to add modal in welcome page. Ugly but does the job for now, especially for the cases where info files are updated (new logins...)

**Possible future developments:**

- separate UMAP and tSNE calculation from these functions, so that UMAP, tSNE are only triggered when really needed. Which could be loading board table clicked (?). This should make welcome page be released quicker, postponing the computation overhead.
- Run the getPGXINFO and getPGXDIR as background computation or assync, as they can take some time.
- Bring the getPGXINFO and `getPGXDIR' even earlier or to a better server spot.

**tests:**

logged, not logged
starting OPG with and without datasets-sigdb.h5, datasets-allFC.csv, datasets-info.csv
adding a pgx in data/

**Please help with tests, as this is a critical module in OPG!**

### DIFF
--- a/components/board.loading/R/loading_server.R
+++ b/components/board.loading/R/loading_server.R
@@ -500,6 +500,8 @@ LoadingBoard <- function(id,
         warning("[LoadingBoard:getPGXINFO] user not logged in!")
         return(NULL)
       }
+
+      pgx.showSmallModal()
       info <- NULL
       pdir <- getPGXDIR()
       shiny::withProgress(message = "Updating library...", value = 0.33, {
@@ -520,6 +522,7 @@ LoadingBoard <- function(id,
       for(s in missing.cols) info[[s]] <- rep(NA,nrow(info))
       ii <- match(info.colnames,colnames(info))
       info <- info[,ii]
+      removeModal(session)
       info
     })
 
--- a/components/board.loading/R/loading_server.R
+++ b/components/board.loading/R/loading_server.R
@@ -607,6 +607,7 @@ LoadingBoard <- function(id,
                     not showing table!")
         return(NULL)
       }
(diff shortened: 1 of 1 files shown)


ivokwee:
we could put the modal or progress bar in the initDatasetFolder() in playbase as an option, or make a detect function datasetFolderNeedUpdate()  that (without actual updating) determines is an update is needed, or not. If needed than show the busy modal, if not, skip.

mauromiguelm:
yep that's the direction I wanted to go, I am working on finishing and testing a PR today and then will get the flickering fixed!

Thanks Ivo!

mauromiguelm:
> we could put the modal or progress bar in the initDatasetFolder() in playbase as an option, or make a detect function datasetFolderNeedUpdate() that (without actual updating) determines is an update is needed, or not. If needed than show the busy modal, if not, skip.

Exactly. `pgx.scanInfoFile` now scans the file, returns TRUE or FALSE, and dending on flag we run `pgx.initDatasetFolder`.

mauromiguelm:
> Main change:
> 
> * remove flickering when the update is fast
> 
> I also want confirmation that we actually want a "wait" modal on the welcome page while this update occurs. It is like the user has to wait on every single page. I would prefer this update to only be triggered on the data loading page, but I know that's tough

fixed

PR #508, opened 2023-06-21
Cross share

ivokwee:
Nick. I changed the sharing flow that the shared file is not temporarily in the data_shared folder with proper from and to postfix. In this way we do not need to create or peek into other user's folders. I think that is less error prone and safer. Also instead of a popup, I have put the notification in a bs_alert box. I like that better. At login, at the load table page, the app will scan the shared_folder with any file that is sent to the user.

One thing you have to fix is that the accept buttons do not work on a second time click (after dismiss). Then you have to clikc on a different received pgx, then cancel, and go back. I think you had a similar problem before. Perhaps the input$ value does not record a change the second time you click. 

Using the shared folder as sharing intermediate, is I think easier to implement for the cross-server sharing.


ivokwee:
I also think we should call the 'Shared' tab the 'Public' tab, and maybe have three folders: data, data_shared, and data_public.

mauromiguelm:
I love it. Great work Nick and Ivo! Will test later today and give feedback if needed.

PR #509, opened 2023-06-22
Separate Public and Shared folder

ivokwee:
Move and rename old data_shared folder to data_public to denote better the difference with data_shared that is now primarily used for user datasharing. The Public folder function now for "public" data.


mauromiguelm:
alright I am investigating, please do not merge

mauromiguelm:
1. `public` datasets that are imported to `user` datasets are not shown in `user` table.. 

image

mauromiguelm:
The "Shared" tab is not showing for me, it would be nice to show and display a message "No one shared a dataset with you yet", so people know about this functionality!

mauromiguelm:
This functionality is just becoming more and more beautiful. Some suggestions:

1. do not show two bslib alerts at the same time, the datasets bs alert is already quite intuitive, people will understand this is the loading page ;)
image

2. bring back requirement to type email twice, this will reduce changes of typing incorrect email by a lot (if typing)

image

ivokwee:
I am going to merge then. The UI changes we can do later.

Issue #507, opened 2023-06-21
Contrasts names remamed with sample names


mauromiguelm:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/mauro/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/mauro/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

Contrast gets renamed based on sample after legacy OPG checks, this is potentially very dangerous. What should I do @ivokwee? I would say user has to fix mismatch... but maybe there a reason behind it

sample_names | contrast_names | contrasts_output
-- | -- | --
notact_004 | notact_004 | notact_004
notact_007 | notact_007 | notact_007
notact_008 | notact_008 | notact_008
notact_009 | notact_009 | notact_009
notact_010 | notact_010 | notact_010
notact_011 | notact_011 | notact_011
notact_014 | notact_014 | notact_014
dsadsad | act12h_018 | dsadsad
act12h_019 | act12h_019 | act12h_019
act12h_020 | act12h_020 | act12h_020
act24h_023 | act24h_023 | act24h_023
act48h_026 | act48h_026 | act48h_026
act48h_027 | act48h_027 | act48h_027
act72h_028 | act72h_028 | act72h_028
act72h_029 | act72h_029 | act72h_029
act72h_030 | act72h_030 | act72h_030
act96h_032 | act96h_032 | act96h_032
act96h_034 | act96h_034 | act96h_034

</body>

</html>

PR #512, opened 2023-06-22
Move all input file checks from OPG to playbase

mauromiguelm:
**WARNING: This requires https://github.com/bigomics/playbase/pull/19**

On this PR I moved all input checks of input files that were in OPG to playbase, as funcional calls..

**Goal:**
-  Allow users (DNAnexus) to check their files before creating pgx externally. Before this was not possible, as these checks were not funcional calls.
- Easy to add new checks and tests.
- Avoid corrupted input files ingestion into omicsplayground/pgx compute
- Allow Nick to show check errors in data preview.

This will close multiple issues, as we avoid corrupted pgx injection into compute pgx. 

For example, this issue is now fixed. https://github.com/bigomics/omicsplayground/issues/507
image

Output of check functions:
- Corrected input files (if possible and safe to do so)
- List of checks and behaviors are here: `playbase::PGX_CHECKS`. Most checks (but not all) are yet implemented, you can search for the error ID in the upload module (e1, e2, e3..). The ones with description are implemented.

Checks I performed:
- Short and long contrast forms
- MANY datasets with and without problems.

Future:
- This will eventually be used by Nick on data preview.
- Expand checks based on testing and _ad hoc_ knowledge.

**This is a critical PR, please help me checking!**

### DIFF
--- a/components/board.upload/R/upload_server.R
+++ b/components/board.upload/R/upload_server.R
@@ -480,8 +480,8 @@ UploadBoard <- function(id,
             })
           }
 
-          uploaded[["samples.csv"]] <- FILES_check$samples
-          uploaded[["counts.csv"]] <- FILES_check$counts
+          uploaded[["samples.csv"]] <- FILES_check$SAMPLES
+          uploaded[["counts.csv"]] <- FILES_check$COUNTS
           samples1 <- FILES_check$SAMPLES
           counts1 <- FILES_check$COUNTS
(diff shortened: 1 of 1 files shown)


mauromiguelm:
> This looks good to me. The only thing I would check is what happens when there are multiple issues. The user sees a modal for the first one, closes it, then sees a modal for the second one, and so on, right? If you've checked this then I think it's OK to merge from my side.

it handles multiples issues!

PR #529, opened 2023-06-28
Simplifying contrast nomenclature: first steps in moving towards comparisons

### LINKED ISSUE #519: switch naming of contrasts to comparisons
switch naming of contrasts to comparisons: comparisons are more familiar to users

mauromiguelm:
**Problem and examples:**

The name contrasts is not intuitive. In many places in the platform, we need to explain what contrasts are. Often, we use comparisons to explain contrasts. Why not use comparisons directly?

image

image

image

**First step:**
1. Replace contrast to comparison in upload board. Now it works with both `contrasts.csv` and `comparisons.csv`.

**Next steps:**
1. If this proposal is accepted, I will update documentation (readthedocs) and App.
2. rename `example data` file contrast.csv to comparisons.csv
3. apply the same concept across boards, where we rename contrasts to comparisons app-wide (now restricted to upload module
4. update playbase to accept both CONTRASTS and COMPARISONS as input, but this is optional.

**Results:**
image

image

this is the fist PR to fix #519

thanks @ncullen93 for helpful discussions.


mauromiguelm:
> This appears to only change labels for the user and not any variables or functions in the underlying implementation, right? Seems fine to merge from my side. I like the idea to use comparisons instead of contrasts.

Correct - minor adjustment to accept `comparisons.csv`

ivokwee:
I am not sure about this. We use the name contrasts everywhere, also inside the pgx object. I am fine using 'comparisons' instead of 'contrasts' in all documentation and in the UI, but to change that everywhere in the code, or renaming files will open up a can of worms. What are your plans?

PR #531, opened 2023-06-28
add setting trigger to NULL to fix unresponsive button after click

ivokwee:
The share buttons were unresponsive after the first click. Setting rl$pgx_share and rl$pgx_share_public to NULL if the trigger value is not needed anymore seems to fix that.


ivokwee:
This is a second fix. Previous commit crashed. Should be good now.

ivokwee:
Sorry need it now. I am merging it.

PR #539, opened 2023-07-02
Fix #428: Wikipathways plots have truncated text in V3

### LINKED ISSUE #428: Wikipathways plots have truncated text in V3 (3.0.RC6.230.426)
I noticed this bug several times: The platform seems to "chomp" the first letter in the Wikipathways plots. Interestingly, it seems to always affect text written in a certain format (see below).

image
image
image

ESCRI11:
This closes #428 

## Description
The initial issue was not on our side, the actual svg's downloaded from WikiPathways have the truncated text.

I have changed the way of displaying the pathways. WikiPathways has an interactive pathway viewer that can be embeded to other website (via an `iframe` on this case). This has pros and cons:

**Pros:**

- We solve the truncated text issue
- Way faster than before. Improved user experience
- We have clickable boxes for further information, which makes the tab more interactive and information-rich (Video 1)
- We can remove the `svg` assets from our repository

**Cons:**

- It requires internet connection to function

**Video 1:**

https://github.com/bigomics/omicsplayground/assets/10220503/924f58ca-9ae8-483b-97f5-acf2a79fc809

Maybe @AxelBigOmics input is required to assess if this enhancement is useful.


ivokwee:
But in our previous version we color the gene by their foldchange. I think that it needed. That is why it also was slow, because it colors the node/genes inside the SVG. I see from the website https://www.wikipathways.org//help.html#widget you can do it using query parameters `?red=ACE&lightgreen=AGT`. Worth to try.

ESCRI11:
Thanks for the heads up! I have noticed that we can also paint them with HEX color codes. So we can do `?efa=ACE&cf3=AGT` and reproduce what we were doing with the svg. I will test if it has any limits when coloring too many cells. I will report back.

ESCRI11:
@ncullen93 Waiting on Ivo approval of the funcionality. Commented out code removed, thanks for the heads up

ivokwee:
OK. good. You can also remove the local SVG and fluctuator stuff. Then proceed with merge. I saw that the code is already behind. color_from_middle seems to be moved from playbase to omicsplayground/ui

PR #552, opened 2023-07-07
add "grey compute" for biomarkers

phisanti:
Solved and tested the Compute issue (see images)
active button
inactive button

Please, @ncullen93 or @ivokwee give it a test


ivokwee:
Also next time the title of the issue should be more descriptive... "grey compute"??

phisanti:
I applied all the suggested changes. I have also changed a bit the logic test so that is a bit more readable.

PR #553, opened 2023-07-07
ENABLE_USER_SHARE prevents shared datasets messages

### LINKED ISSUE #549: `shared datasets` is not hidden when it should
when we options for share is false, we still see the tab

mauromiguelm:
**Summary:**
- `ENABLE_USER_SHARE` now disables all shared dataset messages when set to FALSE.
- Even if user has shared datasets, `ENABLE_USER_SHARE = FALSE` will not show those messages. This is important if user goes from paid to free/demo, where it will not be allowed to share datasets anymore.

**Possible future improvements:**
- Hide Sharing tab if no datasets available OR if `ENABLE_USER_SHARE = FALSE`. This was the initial idea, but its nice to have the sharing tab with the message, so user knows the functionality exists but it is disabled.

**Tested with combinations of the following conditions:**
- ENABLE_USER_SHARE set to FALSE or TRUE
- With and without shared datasets

this fixes #549

### DIFF
--- a/components/board.loading/R/loading_server.R
+++ b/components/board.loading/R/loading_server.R
@@ -64,6 +64,11 @@ LoadingBoard <- function(id,
     )
 
     if (enable_user_share==FALSE) {
+      
+      output$sharing_alert <- renderUI({
+        bs_alert(HTML("This table shows the <b>available datasets</b> in your library. The table reports a brief description of each dataset. The <b>Signature t-SNE</b> shows similarity clustering of fold-change signatures using t-SNE. Select a dataset in the table and load the data by clicking the <b>Load Dataset</b> button below."))
+      })
+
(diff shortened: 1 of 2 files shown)


mauromiguelm:
> I think you mean that you tested with ENABLE_USER_SHARE and not public, as changing ENABLE_PUBLIC_SHARE has no effect. I think this is good, but the only issue I see is that the bs alert on the User tab is for some reason missing when ENABLE_USER_SHARE is false.
> 
> See with and without:
> 
> Screen Shot 2023-07-07 at 11 35 32 AM Screen Shot 2023-07-07 at 11 36 41 AM

good catch Nick, fixed!

mauromiguelm (commented):
fixed in 881201e1

Issue #492, opened 2023-06-17
plotmodule plotoptions dropdown is clipped within card window


ivokwee:
The plot options dropdown is clipped to the card window and causes the scrollbars to appear. We had this before and solved it somehow with help of Carson but now it appears again. Using some setting it should be possible to let the dropdown float and overflow outside the card borders.

![image](https://github.com/bigomics/omicsplayground/assets/19613146/ba8d43fa-0b77-40fa-876d-66b8afd43495)

ESCRI11:
Looks to be related to the use of `bslib::layout_column_wrap`. Elements placed using traditional `col-md-X` divs do not display this behaviour (e.g. Load dataset tab)

<img width="225" alt="image" src="https://github.com/bigomics/omicsplayground/assets/10220503/524cbb11-27ca-4afd-9ba0-e9afefe83435">

Confirmed by moving elements out of `bslib::layout_column_wrap` and they automatically behave OK.

mauromiguelm:
on it, I fixed this problem some months ago, not sure why it came back.

Investigating.

cpsievert:
I'd be surprised if the new version of `{bslib}` did, in fact, introduce this problem. That said, with both the older and newer version of `{bslib}`, it is indeed true that, if you put a `card(style = "overflow: visible")`  in a parent element that has `overflow:auto` set (e.g., `layout_column_wrap()`) you'll run into this issue (because the parent's `overflow` setting will overrule the child's property). Here's a minimal reprex of the problem (notice how the `selectInput()`'s in the 1st row get clipped but the 2nd doesn't):

```r
library(shiny)
library(bslib)

my_card <- function() {
  card(
    style = "overflow:visible;",
    card_header(
      selectInput("foo", NULL, state.name),
      class = "d-flex justify-content-end"
    ),
    "Body"
  )
}

ui <- page_fluid(
  layout_column_wrap(
    width = 1/2,
    my_card(),
    my_card()
  ),
  my_card()
)

shinyApp(ui, function(...){})
```

You can currently work around the problem by making sure the layout containers also have `overflow: visible`, but it's definitely on my (mental) TODO list to make this sort of thing easier...

```r
layout_column_wrap_visible <- function(...) {
  res <- layout_column_wrap(..., style = "overflow:visible;")
  tagQuery(res)$children()$addAttrs(style = "overflow:visible;")$allTags()
}

ui <- page_fluid(
  layout_column_wrap_visible(
    width = 1/2,
[...]

PR #554, opened 2023-07-07
fix for dropdown overflow issue #492

ivokwee:
This fixes the overflow bug #492 that the dropdown was clipped within the card borders and could cause overflow (adding scroll bars) to the card. It now overflows outside the card. 

**Before:**
image

**After:**
image


ivokwee:
The Dropdown implementation is in ui/ui-DropDownMenu.R   I wonder if the bsutils::Dropdown could be used instead. Something for later.

PR #555, opened 2023-07-07
fix sharing email lower/upper case mismatch

ivokwee:
From Murat:

In data sharing, do email addresses and usernames have to match exactly when exchanging data? I mean, is the matching case sensitive?

For example, here is a real case example: Alice.PEO1v_PEO4.all.pgx__to__Alice.filipe@*******__from__blstccs@******* where Cristina shared a dataset with Alice. She entered Alice.filipe@****** but Alice's account is alice.filipe@****** Does she get a notification and the dataset?

**FIXED**
- sender email is converted to lowercase before sending
- received files are matched with 'ignore.case=TRUE'


ivokwee:
test receiving:

PR #565, opened 2023-07-09
User trace. Do not allow concurrent sessions

ESCRI11:
## Description
With the current implementation of the login, there is no check to block concurrent sessions using the same credentials. That can result in clients being able to use more seats than contracted.

In order to control this issue, a simple check is performed. This check is based on user heartbeat and user login/logout trace.

## Recording the trace
The user trace is recorded on a file under the `trace` folder by the name of `trace_log.txt` (folder is on `.gitignore` to avoid leaks on GitHub repo, it is created automatically if missing [here](https://github.com/bigomics/omicsplayground/commit/9d29f55453e4ea76edddfba51d89843e2d2017a7)).

The recorded trace lines have the following structure:
`email` \t `heartbeat/login/logout` \t `POSIXct timestamp`

Therefore, this method only works when using password based authentication, which in the case of OPG that can be applied to regular `password` authentication or `firebase`. Please note that `firebase` is not yet implemented, waiting the `password` method be reviewed and approved before that. It should be very trivial to support `playbase`, just a few changes to `AuthenticationModule.R` will do.

## Current status
With the proposed changes, the user will not be allowed to login if their account is already active. This is checked using a two minute heartbeat, if the heartbeat is not older than two minutes (or the user has not logged out using the logout button), the login will not be successful. 

## Problems
There is just small edge case that is problematic. If the user for some reason refreshes the application (F5), there will not be a logout record and the heartbeat will be newer than two minutes, therefore the user will have to wait for two minutes (or less) before being able to log in again.

## Future upgrades

- `firebase` implementation
- `trace_log.txt` file cleaning. Maybe we do not need all time tracing. E.g. remove traces older than 1 month


ESCRI11:
Agreed. I will work on faster options + async with promises. 

Not sure how often it does happen, it was a request from Ivo to have this implemented. We could log the incidents easily.

ESCRI11:
@ncullen93 What are your views on using a local SQLite database? This allows for queries using `DBI` package so only the required lines are loaded into R + `promises` async implementation? That way in the future the database can be moved outside of the local folder and point all deployments to it.

ivokwee:
For the chatbox I implemented something with fread/fwrite that was almost as fast as DBI. Still DBI was faster. But I think to be more efficient we should separate login logging and lockfile, so the login/logout log can be written slow S3 (for us easy to maintain across servers), while the lock file can be in the userfolder with just a single one line file, or even empty file. I am l looking a this now.

ivokwee:
I rewrote the code by moving the checking in the app.R so it should work for all login methods. However it activates only after login, instead at the login prompt. I think that is perhaps OK so the "disappointment" is bigger. For the lock, it writes an empty file in the user folder with the username and session in the filename. A super small file should be fast. The user access logging is done in a separate file in the etc/access.log  file, It is written at login, explicit logout, or when detecting a stale logout.

ivokwee:
It's pretty light weight. We can adjust the poll time (heartbeat) which only affects the blocked user waiting. I have added a remove lock at session end, so in case of a clean logout there is no lock file. Also the code is now R6 and very clean in the app.R. Going to merge.

PR #580, opened 2023-07-15
Fix processx pgx folder not matching new login user folder

mauromiguelm:
- each raw__ folder has the original pgx path saved as a txt file (pgx_folder.txt).
- if the path in pgx_folder.txt does not match auth$options, then we do not load pgx.
- pgx will be saved in user_dir folder, so it has no contact with new user

its a solution that will increase security, the next step is to find a way to clear upload module between loging on the same session.


ivokwee:
I simplified few things. I guess the save_folder.txt file is not needed as you can read the rds file.  

The strange thing now is that the PGX file is actually written twice, once in the background process, and then it is read in the app, and passed to the uploaded_pgx reactive, then get initialized and save again here https://github.com/bigomics/omicsplayground/blob/7419f269bb65144f54236f76ac4cfc439fc98543/components/board.upload/R/upload_server.R#L74C6-L87C18 . Perhaps the second save can be skipped.

mauromiguelm:
I really liked the changes, could not spot any problems!

Feel free to merge!

PR #582, opened 2023-07-16
Fix #559: Uploaded pgx files dont appear immediately in data loading table if they are loaded

### LINKED ISSUE #559: Uploaded pgx files dont appear immediately in data loading table if they are loaded
If you upload a pgx file and then decide to load it immediately (taking you to the dataview page), you wont see the dataset in the data loading table if you go to the data loading page. The dataset will show up if you resize the page and it will show up immediately the next time you log in. I fixed this problem somehow previously but now it is back. Perhaps the solution would be to use DTproxy to update the data in the datatables.

ESCRI11:
This closes #559 

## Description
Selecting the tab and immediately afterwards loading the data made the reactivity not work. That is why selecting "Stay here." worked ok.

Just loading directly the data fixes the issue.

Please @ncullen93 double check and merge.

### DIFF
--- a/components/board.upload/R/upload_server.R
+++ b/components/board.upload/R/upload_server.R
@@ -104,8 +104,7 @@ UploadBoard <- function(id,
 
       load_my_dataset <- function() {
         if (input$confirmload) {
-          bigdash.selectTab(session, selected = "load-tab")
-          load_uploaded_data(new_pgx$name)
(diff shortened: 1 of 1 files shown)


ivokwee:
not sure. I never had such problem.

ivokwee:
Not sure. Both cases work for me. As the second case is more minimal (this PR), less code is good. mergin.

PR #587, opened 2023-07-18
Carson fix dropdown menu v3

ivokwee:
Attempt number three of Carson's fix of the dropdown menu. Previously the dropdown got clipped inside the card and caused overflow (scrollbars). With this fix, the dropdown goes outside the card.


ESCRI11 (approved):

ESCRI11:
It works on my end, but there is a noticeable lag between clicking and the action taking place.

PR #588, opened 2023-07-18
feat: remove `fluctuator` dependency

ESCRI11:
## Description
With the upgrade of WikiPathway plots the `fluctuator` package was not really required at all. However, there were still some traces of it left.

Removed them to have one less dependency.


ivokwee:
OK. one package less...  now 974 packages to go...

PR #606, opened 2023-07-28
Dev card clipping (JS function)

ivokwee:
Xavier solution for dropdown clipping problem using a JS function. The JS code is now coded as function to prevent too much code attaching to each card/plotmodule.  Previous solutions still crashed some browsers with out of memory error. Please test and report.

### DIFF
--- a/components/ui/ui-DropDownMenu.R
+++ b/components/ui/ui-DropDownMenu.R
@@ -102,13 +102,6 @@ $('#", id, "').on('click', function(){
 ") ## end of paste0
 }
 
-dropdown.assets <- function(id) {
-  tagList(
-    shiny::singleton(tags$style(HTML(paste0("#", id, " .dropdown-menu { display: none; }")))),
-    tags$script(HTML(dropdown.jsCode(id)))
-  )
-}
-
 DropdownMenu <- function(..., size = "default", status = "default", icon = NULL, width = "auto", margin = "10px") {
   id <- bigdash:::make_id()
   tags$div(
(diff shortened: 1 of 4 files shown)


ESCRI11 (approved):

components/ui/ui-DropDownMenu.R
```diff
@@ -183,7 +77,6 @@ actionMenu <- function(..., size = "default", status = "default", icon = NULL, m
         },
         error = function(w) {}
       )
-    ),
-    dropdown.assets(id)
+    )
```
cpsievert, line 80: Are we sure this isn't also needed for `ActionMenu()`?

cpsievert (approved):
LGTM (other than the one question)

components/ui/ui-DropDownMenu.R
ESCRI11: Good catch. Just fixed it

phisanti:
I have just tested this new version, and I can confirm that the "Out of memory error" does not occur. However, I see that there are 3 versions of the JS code: `dropdown.jsCode.SAVE2`, `dropdown.jsCode.SAVE1` and `dropdown.jsCode`. Before merging into the main, these should be clean so we don't clutter the main project. Ideally, we should consider moving the `www` directory.

ESCRI11:
@phisanti It is removed here https://github.com/bigomics/omicsplayground/pull/606/commits/a469f4d88a57701af51248dc379662ed82afc086

PR #611, opened 2023-08-03
Warn user and CS user when pgx compute process fails

mauromiguelm:
**Summary:**
- **When a pgx fails**, the on_process_error is called. Previously no warn to the user was given.
- We have many pgx computation that triggers on_process_error, even when no clear error is observed. We will need to debbug these cases (Ivo knows a few examples).
- The combination of the two points above created a lot of confusion and frustation to the user.
- **For the user**, I added a modal warning with error log and always send an email to the user when auth$email is given.
- **For the customer support**, the user has the option to give the data and logs to the customer support by clicking the "send data to customer support" in the modal.

**Next steps:**
- send email to user when dataset is completed

@phisanti we can go through it so you can a feeling for how we handle background computation!
thanks @ncullen93 for solving integration of blastula with observers, it was very easy to implement this!

**modal pop up:**
image

**confirmation and close of previous popup:**
image

**Message to user:**
image

**Message to customer support (support@bigomics.ch)**
image


phisanti:
Yeah, let me know if you have time tomorrow @mauromiguelm

Issue #613, opened 2023-08-03
Add option (default) to have an independent scaling of y-axis on volcano grids


mauromiguelm:
**Summary:** 

1. Add option (default) to have an independent scaling of y-axis, and not a single scaling across all volcano grids, as sometimes there are big differences between statistics which makes impossible to see results.

2. In v2, we had independent scaling for each plot. We should have both options, the user should choose.

3. these types of plots (volcano grids) are present on diffexpr and genesets, so be sure not to miss any.

<img width="761" alt="image" src="https://github.com/bigomics/omicsplayground/assets/28757711/22eee2cd-c067-4326-bee7-7e22dc0f02bf">

phisanti:
This should be an easyone. I will try to get it done today.

PR #616, opened 2023-08-05
add scale per method option, set default=TRUE

phisanti:
Add some minor changes addressing #613. Basically I added:

```r
  plot_options <- shiny::tagList(
    withTooltip(shiny::checkboxInput(ns("scale_per_method"), "scale per method", TRUE),
      "Scale the volcano plots individually per method..",
      placement = "right", options = list(container = "body")
    )
  )
```

and 

```r
          if (input$scale_per_method) {
            ymax1 <- 1.2 * quantile(xy[, 2], probs = 0.999, na.rm = TRUE)[1] ## y-axis
          }
```

To the files `enrichment_plot_volcanoall.R`, `enrichment_plot_volcanomethods.R`, `expression_plot_volcanoAll.R`, and `expression_plot_volcanoMethods.R`. 

I would like to note that there is space for improvement as the plot re-renders each time that the user expands the options panel, regardless of whether they click or not on the option. Any suggestions will be welcomed. 

I set @mauromiguelm as reviewer since you opened the issue.

### DIFF
--- a/components/board.enrichment/R/enrichment_plot_volcanoall.R
+++ b/components/board.enrichment/R/enrichment_plot_volcanoall.R
@@ -12,12 +12,20 @@ enrichment_plot_volcanoall_ui <- function(
     width) {
   ns <- shiny::NS(id)
 
+  plot_options <- shiny::tagList(
+    withTooltip(shiny::checkboxInput(ns("scale_per_method"), "scale per method", TRUE),
+      "Scale the volcano plots individually per method..",
+      placement = "right", options = list(container = "body")
+    )
+  )
+
   PlotModuleUI(
     ns("plot"),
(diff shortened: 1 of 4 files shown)


mauromiguelm (changes requested):
**problem in diff expr > volcano all** 
image
**- diff expr > volcano methods has no scaling**

**- geneset board volcano all and methods seems to work fine**

phisanti:
I have added:
```r
      xmax <- max(1, 1.2 * quantile(abs(unlist(pd[["F"]])), probs = 0.999, na.rm = TRUE)[1]) ## x-axis
```
to `expression_plot_volcanoAll.R`. That should fix it.

mauromiguelm (approved):

PR #617, opened 2023-08-05
feat: email user when pgx is completed

mauromiguelm:
somehow this code got lost, since Santhiago accepted the PR, I assume checks were performed. Merging now.


mauromiguelm:
since I tested this extensively, and likely Santhiago as well, I will merge @phisanti @ivokwee

PR #626, opened 2023-08-14
Notify logged in user when a new dataset is received

### LINKED ISSUE #624: When user is logged in and receives a dataset, automatically notify user
**problem:** if the user is logged in and receives a shared dataset, he will have to log out and log in again to be able to see the share dataset.

**solution:** detect new datasets and notify user as soon as a new share dataset is received.

mauromiguelm:
this closes #624 

**Summary:**
- added a memmory of how many datasets user has: `nr_ds_received`
- if reactivePoll detects that `current_ds_received` is higher than the `nr_ds_received`, we show the modal warning the user
- simplified reactive logic by removing dependence on `current_page` and `refresh_table`, now relying only on reactivePoll

**Result:**

- When we drop a dataset in the shared folder, we get a notification within 10s (reactivePoll timer set to 10s).
image


ESCRI11:
Similar to PGX compute, it would be nice to have a "Take me there" button and a "Stay here" instead of just "OK".

image

components/board.loading/R/loading_module_received.R
```diff
@@ -17,38 +17,81 @@ upload_module_received_server <- function(id,
...
-      ## ------------ get received files
-      getReceivedFiles <- shiny::reactive({
-        req(auth$logged)
-        if (!auth$logged) {
-          return(c())
+      # function that listed to input new_dataset_received
+      show_shared_tab <- function(){
+        if (input$new_dataset_received) {
+          bigdash.selectTab(session, "load-tab")
+          shinyjs::runjs('$("[data-value=\'Sharing\']").click();')
```
mauromiguelm, line 26: @ESCRI11 I had a few crashes with this line before, since there may be situations where UI was rendered when the js line is run, and the app crashed. I tried several solutions (on document ready, only run if Sharing exists) but no good solution.

Since we have a message in yellow telling the user that we have shared datasets, I opted to not redirect to Sharing!

I could not reproduce (welcome page, load example data > introduce shared dataset > click to go to shared tab), and it worked.. not sure why its working now, but if it does not fail again, lets merge!

T

Issue #548, opened 2023-07-06
Adding "exclude" option or multiple selection for genesets annotation


AxelBigOmics:
This is more a request than a bug, but I noticed that by expanding the pathways and genesets database, a lot of the informative pathway get "drowned" by non-descriptive gene silencing or upregulation experiments. This is particularly noticeable when I select "all" under the "Geneset UMAP".

While it is great to have more datasets for comparative analysis, I would appreciate the option of excluding some of them (e.g. LNCHUB, RNASEQ, BIOPLEX), while leaving the rest. This is because the more thoroughly descriptive HALLMARK/Msig/etc... pathways are more informative when initially exploring the data. 

In fact, retaining the dataset collections that were present in V2 without the new ones ( like a new "all minus"option) would be great. Would it be possible to add such an option in the scroll down menu?

![image](https://github.com/bigomics/omicsplayground/assets/85561624/c760f749-daae-4b8d-866a-1dfab113d645)

ivokwee:
Initally I thought this is confusing "all minus"?? minus what? I think Axel means a "exclude" function. We could either allow multiple selection (which I prefer more), perhaps even in the global settings page.

PR #629, opened 2023-08-15
Add multiple choices gene and geneset annotation

ShaliniPandurangan:
## Description

Enable multiple choice = TRUE in the annotate genesets and annotate genes under the board featuremap. 
## Related Issue
 #548


mauromiguelm (approved):
- since we use validate, we did not need the `shiny::istruthy`, so I removed it
- apply is more efficient than for loops, so I replaced them for apply
- since Axel asked for an option to exclude genes families/genesets, one easy way to accomplish this is to allow multiple inputs in these selectInput calls. So I added that, and the user can select their custom families/pathways of interest
- simplified code

@ShaliniPandurangan can you test once again?

ShaliniPandurangan:
Thanks for the changes. I tested it again, it works fine.

PR #645, opened 2023-08-24
Fix #620: improve pgx computation spinner

### LINKED ISSUE #620: improve pgx computation spinner
**problem:** a few users complained that the spinner is not visible enough and they do not notice that the computation is running.

**solutions:** Maybe we can add a log widget on the top bar (similar to chirp), where the main activities of the user are tracked and every time a pgx is uploaded we have a "new" message . Or simply make the widget more noticeable.

image

ESCRI11:
This closes #620 + header refresh (still not tracked on issues, only commented on #620 discussion by @mauromiguelm)

### Header refresh
The header refresh was being triggered by [this line](https://github.com/bigomics/omicsplayground/compare/fix-%23620?expand=1#diff-a03e6d6a363b07de38c96f79f8b756c0326a33d8ad73ab1033b705d344830b1cL449) (reactivity). Moved the call to a reactive value so that the header only is refreshed when we intend it to.
### #620 fix
The spinner itself (IMHO) is pretty ok, I don't really think we have to make it more visible or move it somewhere else. However we kept loosing it on header refresh, which is fixed by this PR. Another edge case that made it disappear :

1. Creating compute job
2. Loading dataset

On this case the header does indeed have to be refreshed to display the loaded dataset name, by sharing the `input$nav` from the server to `upload_module_computepgx.R` I trigger [this code](https://github.com/bigomics/omicsplayground/compare/fix-%23620?expand=1#diff-d2a224220950205ab698eac5e55a59807e45eaafad1db8ae096d59d4316fc5e2R729-R746) at each tab change, therefore we will always display properly if a compute job is running.

### DIFF
--- a/components/app/R/server.R
+++ b/components/app/R/server.R
@@ -146,6 +146,7 @@ app_server <- function(input, output, session) {
       pgx_dir = PGX.DIR,
       pgx = PGX,
       auth = auth,
+      nav = reactive(input$nav),
       reload_pgxdir = reload_pgxdir,
       load_uploaded_data = load_uploaded_data
     )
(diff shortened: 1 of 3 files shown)


mauromiguelm (approved):
- spinner was being added the container (beforeEnd), meaning that when dataset updated, spinner gets lost. this was fixed by using beforeBegin, so spinner is outside .current_dataset container.
- removed nav from upload board

PR #646, opened 2023-08-24
Fix #621: Add a separate column in tables that has link to genes/proteins

### LINKED ISSUE #621: Add a separate column in tables that has link to genes/proteins
**Feedback from user:** The hyperlinks in the tables (such as diff expression) need to be placed separately. not as hyperlink in the text, but maybe with an arrow button at the end of the text.

ESCRI11:
This closes #621 

## Description
Following @ivokwee comments on #621 issue conversation, the only table that was tedious to use was `Expression > Differential Expression > Gene sets with gene`. Nevertheless, I think it made no sense placing separate column on that one and leave other tables with text links, feeling should be as consistent as possible. Therefore I have modified all tables that had text links.

To find all the tables to be changed I looked at the uses of `wrapHyperLink`, which pointed 7 different tables using text links.

I have placed icons as rownames with the links to the correspondent sites. (Figure 1)

On the old implementation, cases with no link found had a different font color (compare Fig 2 and 3, personally I struggle to identify Fig 2 missing links on first sight); with the new developments, since we use icons, some extra step was required: I overplace two icons to visually identify that no link is available. (Figure 3)

### Figure 1
image

### Figure 2
image

### Figure 3
image


mauromiguelm (changes requested):
beautiful! We are missing a few tables

geneset > pathway analysis > wikipathways > Enrichment table (https://github.com/bigomics/omicsplayground/blob/9145a92ee8dc4caf11ebb51f4ac7803a69cdc738/components/board.pathway/R/functional_table_wikipathway.R#L64-L69)

geneset > pathway analysis > reactome > Enrichment table 
https://github.com/bigomics/omicsplayground/blob/9145a92ee8dc4caf11ebb51f4ac7803a69cdc738/components/board.pathway/R/functional_table_reactome.R#L50-L54

mauromiguelm (approved):

PR #654, opened 2023-08-29
Fix #649: Enrichment analysis has first column truncated

### LINKED ISSUE #649: table  Genesets > Geneset enrichment > Enrichment analysis has first column truncated
table  Genesets > Geneset enrichment > Enrichment analysis has first column truncated

ESCRI11:
This closes #649 

## Description
The handling of the rownames was being made inside the table render function and not on the `table_data` generator. Moved one line to achieve that and have the actual geneset names on the downloaded csv.


ivokwee:
Before fix:
image


After fix:
image

PR #656, opened 2023-08-29
Upload preview v2

ncullen93:
This is an updated version of the data upload preview. It should work well now. An upgrade is that now, if a file errors then it will still show the file in the preview with the errors. The errored file will not be actually uploaded to the platform though.


mauromiguelm:
Hi Nick, it was data that passed all tests before, so its definitely strange.. I will text you on slack

mauromiguelm:
it works quite nice, some feedback:

- we will need to review the pgx checks
- would be nice to add a legend: what green, yellow or red means, so the user will know that red wont be accepted and input will be removed... eventually might need rethink on this logic
- are the logic of pgx checks the same as before?
- we should release this to a few power users and get feedback! @ivokwee?

PR #694, opened 2023-09-18
Fix #637: Tables downloaded as .htm in production (not locally)

### LINKED ISSUE #637: Tables downloaded as .htm in production (not locally)
Some tables are being downloaded as .htm instead of .csv. One example is the table GeneSets > GeneSet enrichment > enrichment(all), which is downloaded as .htm in free.playground and as csv locally.

Its likely a js bug.

ESCRI11:
This closes #637 

## Description
As described on the issue thread, there is an initial link (downloads htm) and a correct one (downloads image/table). As it turns out is not a matter of interacting with other elements of the popover, it just takes some time which seems to be related with client computing browser. 

By disabling the buttons by default and enabling them once the correct href is available, we make sure the download is always reliable.

From my testing it does not affect the usability or experience of the application.

### DIFF
--- a/components/app/R/www/temp.js
+++ b/components/app/R/www/temp.js
@@ -434,3 +434,39 @@ $(document).ready(function() {
 
 
 
+document.addEventListener('shown.bs.popover', function(event) {
+	// Get the popover container (Bootstrap 5 attaches the popover to the body by default)
+	var popover = document.body.querySelector('.popover');
+  
+	if (!popover) {
+	  console.log("Popover not found.");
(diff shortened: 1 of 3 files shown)


mauromiguelm:
thinking if there's a more efficient way to do this - and this case only happens when server is slow, does it affect other UI components as well?

ESCRI11:
Have not found it to affect other components.

mauromiguelm:
and I found one case where the removing class "disable" is not working

image

PR #704, opened 2023-10-02
Feature #696: auto logout user after idle period

### LINKED ISSUE #696: Feature: auto logout user after idle period
**Problem**
A logged in user with an open tab in the browser will stay logged in even a whole day. This is not efficient  (memory usaage) and may block new users to log in because we restrict the number of sessions.

**Possible solution**
Would be good to automatically logout the (logged in) user after say 1hour idle. For logged out users (login page) this is already done using a timer. For logged in users, this is a bit tricky as we may need to monitor any user interaction: any clicks or mouse moves. Not sure if this can be done server side, or has to be done client-side using JS.

ESCRI11:
This closes #696 

## Description
Added a reactive value that acts as an upward counter when there is no activity, we define activity as a user click on the UI. When there is a click, the counter gets back to 0. With this counter we can check the time of inactivity and end the session.

In order to check whether there is a click by the user, I added JS in `bigdash` that triggers the `input$userActivity` observer. https://github.com/bigomics/bigdash/pull/23


ivokwee:
As said. Better take inactivity out of heartbeat because heartbeat (login control) can be disabled in OPTIONS and we want still inactivity tracker active. Otherwise should be fine.

ESCRI11:
Inactivity trigger is now isolated outside of heartbeat, also I added an OPTIONS so it can be toggled for flexibility. @mauromiguelm Can you provide some input on how to avoid ending the session when PGX is being computed?

mauromiguelm:
> Inactivity trigger is now isolated outside of heartbeat, also I added an OPTIONS so it can be toggled for flexibility. @mauromiguelm Can you provide some input on how to avoid ending the session when PGX is being computed?

can we reset the timer every time this reactive is triggered? https://github.com/bigomics/omicsplayground/blob/6594a4af239320acee40daeb44496b103aaf213d/components/board.upload/R/upload_module_computepgx.R#L560

ESCRI11:
@mauromiguelm Implemented on last commit 9fe2cd6, how does it look?

mauromiguelm:
> > Inactivity trigger is now isolated outside of heartbeat, also I added an OPTIONS so it can be toggled for flexibility. @mauromiguelm Can you provide some input on how to avoid ending the session when PGX is being computed?
> 
> can we reset the timer every time this reactive is triggered?
> 
> https://github.com/bigomics/omicsplayground/blob/6594a4af239320acee40daeb44496b103aaf213d/components/board.upload/R/upload_module_computepgx.R#L560

approved from my side!

mauromiguelm:
> @mauromiguelm Implemented on last commit [9fe2cd6](https://github.com/bigomics/omicsplayground/commit/9fe2cd60a373f82cdc239f06b9231f355dd6ccef), how does it look?

approved from my side!

PR #720, opened 2023-10-21
feat: hover dropdown implemented

### LINKED ISSUE #714: bring back "show legends" button from settings tab to the front page, to show plot captions
Today is very difficult to find that button, we should bring it back to the front page.

Nick tried to place it under settings, to have more settings available, but its too hidden. That means users dont get the extra info that comes with plot caption.

ESCRI11:
This closes #714 

## Description
Following #714 discussion, added a hover icon. To test, make sure to have the latest version of `bigdash` installed.

Before merging: 
1. Is the icon OK @ivokwee?
2. Should "App settings" be renamed? There are no settings on there anymore (pic for reference)

image

### DIFF
--- a/components/app/R/server.R
+++ b/components/app/R/server.R
@@ -125,11 +125,16 @@ app_server <- function(input, output, session) {
     nav_count = reactive(nav$count)
   )
 
-  env$user_settings <- UserSettingsBoard(
+  UserSettingsBoard(
     "user_settings",
     auth = auth,
     pgx = PGX
   )
+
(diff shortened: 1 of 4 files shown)


mauromiguelm:
I have a problem, bslib@mains installed fresh.. anything I can do to fix it?

mauromiguelm:
worked but cache has to be cleared

ESCRI11:
- [ ] hover area a bit wider
- [ ] make a wider div holding the settings icon (maybe)

PR #728, opened 2023-11-01
Info button

### LINKED ISSUE #711: give more information about what type of statistical method we have at Upload > compute > options
it would be nice to classify the methods by type, or even add tooltip explaining what they do

ESCRI11:
This closes #711 

## Description
Following the disucssion, added an `info` button (pic for reference) which links to https://omicsplayground.readthedocs.io/en/latest/methods/. Is this OK or extra things shoul be needed to be placed?

image


mauromiguelm:
we would need to rewrite read the docs, so each box has a separate url. we need to split this... 

https://omicsplayground.readthedocs.io/en/latest/methods/#statistical-testing

PR #731, opened 2023-11-06
Fix #727: wgcna > intramodular > Membership-trait heatmap: covariance option not working

### LINKED ISSUE #727: wgcna > intramodular > Membership-trait heatmap: covariance option not working
It should be removed or fixed

ESCRI11:
This closes #727 

## Description
@mauromiguelm Can you check if method is correct? I added the option to calculate the covariance instead of the correlation.

Let me know.


ivokwee (approved):
ok. perhaps the rho3 should be cor? not sure. leave it for now.

PR #764, opened 2023-11-22
Fix `bslib` version

ESCRI11:
## Description
Docker builds use the master branch of `bslib`, therefore when building new images the latest `master` commit is installed. In general this is no problem, however, recently the boostrap version has changed from 5.2.2 to 5.3.1.

This boostrap version change altered some CSS defaults of the application, which changed the actual feel of the application from the development stage to production.

It is important to fix the version, so that this CSS changes are actually audited before making it to production.


mauromiguelm:
nice! I would bring this to docker-base, so we save some time installing it every time

ESCRI11:
Agreed. I moved `bslib` to the base image.

Also, I have removed some packages from the Dockerfile, as they are packages that we pretty much never work on (or are not ours) and were installed at the base already.

PR #803, opened 2023-12-14
New hide and show upload tab logic; show selected species in compute tab; small refactoring

mauromiguelm:
This pull request includes changes to hide and show the upload tab logic at server start, on load example and proceed_to_upload reactive. It also includes a UI refactor and add selected species to Compute tab.

image


phisanti (approved):
Works locally

PR #810, opened 2023-12-14
Fix #802: display correct gene families on "Gene family:" input

### LINKED ISSUE #802: Error when working on the clustering tab
Error when working on the clustering tab if "split sample by gene" and then select any Gene Family except "<all>".

image

ESCRI11:
This closes #802 

## Description
Display only the gene families that contain the gene selected as "split samples by", otherwise, there are some gene families that may not have this gene and the plots will fail

(Removed "fix: drop=FALSE" commit that slipped by from other PR)


phisanti:
The solution looks good, however, have you checked if it will work with other species?

```r
        if (input$hm_splitby == "gene") {
          find_gene <- function(x) {
            any(grepl(input$hm_splitvar, x))
          }
          choices <- names(pgx$families[sapply(pgx$families, find_gene)])
        }

```        
   Sometimes families comes with the human gene name while the gene symbol might come with the local species name. I tried to checkout to fix-#802 but cannot see the branch¿¿

mauromiguelm:
@ESCRI11 

- <contrast> still crashes with dataset below
- gene families are lost, needs to be reviewed with collapse by genes = TRUE and FALSE, and make sure we use orthologs


with yeast dataset (see github issue)

PR #819, opened 2023-12-18
Allow selection of genes within genesets in expression volcano; improve messages to user in case gene has no geneset

mauromiguelm:
Allow selection of genes within genesets in expression volcano; improve messages to user in case gene has no geneset


phisanti:
Tested on _C elegans_ with gene collapsed and _S cerevisiae_ no collapsed. Works in both. Ready to merge.

PR #828, opened 2023-12-21
Fix input_genelistUP bug in signature_server.R

mauromiguelm:
This pull request fixes a bug in the `signature_server.R` file where the `input_genelistUP` function was not working correctly. The bug caused an incorrect match of input genes with probes, resulting in unexpected behavior. The fix ensures that the correct matching is performed.

### DIFF
--- a/components/board.signature/R/signature_server.R
+++ b/components/board.signature/R/signature_server.R
@@ -93,6 +93,7 @@ SignatureBoard <- function(id, pgx, selected_gxmethods = reactive(colnames(pgx$g
 
     input_genelistUP <- shiny::reactive({
       shiny::req(input$genelistUP)
+
       gg <- input$genelistUP
       gg <- strsplit(as.character(gg), split = "[, \n\t]")[[1]]
       if (length(gg) == 1 && gg[1] != "") gg <- c(gg, gg) ## hack to allow single gene....
@@ -106,7 +107,6 @@ SignatureBoard <- function(id, pgx, selected_gxmethods = reactive(colnames(pgx$g
 
       ## Get current selection of markers/genes
       type <- input$type
-
       level <- "gene"
       features <- toupper(pgx$genes$gene_name)
       xfeatures <- toupper(pgx$genes[rownames(pgx$X), "gene_name"])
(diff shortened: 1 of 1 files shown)


phisanti:
For this fix, in file `components/board.signature/R/signature_server.R#140`, I would suggest to use:
```R
        match_input_genes <- intersect(input_genelistUP(), probes)
        if (length(match_input_genes) == 0) {
          # use human ortholog in case no matched found in proves
          probes <- pgx$genes[pgx$genes$human_ortholog %in% input_genelistUP(), "gene_name"]
        } else {
          probes <- match_input_genes
        }
```

because it allows for partial matches within the list. Rather than:


```R
        match_input_genes <- !all(input_genelistUP() %in% probes)

        if (match_input_genes == FALSE) {
          # use human ortholog in case no matched found in proves
          probes <- pgx$genes[pgx$genes$human_ortholog %in% input_genelistUP(), "gene_name"]
        }
        genes <- pgx$genes[probes, "gene_name"]

```

mauromiguelm:
> For this fix, in file `components/board.signature/R/signature_server.R#140`, I would suggest to use:
> 
> ```r
>         match_input_genes <- intersect(input_genelistUP(), probes)
>         if (length(match_input_genes) == 0) {
>           # use human ortholog in case no matched found in proves
>           probes <- pgx$genes[pgx$genes$human_ortholog %in% input_genelistUP(), "gene_name"]
>         } else {
>           probes <- match_input_genes
>         }
> ```
> 
> because it allows for partial matches within the list. Rather than:
> 
> ```r
>         match_input_genes <- !all(input_genelistUP() %in% probes)
> 
>         if (match_input_genes == FALSE) {
>           # use human ortholog in case no matched found in proves
>           probes <- pgx$genes[pgx$genes$human_ortholog %in% input_genelistUP(), "gene_name"]
>         }
>         genes <- pgx$genes[probes, "gene_name"]
> ```

This is the current logic in this PR. Before if we had any element not matched, it would use ortholog. This crashed for human, as we can always input genes that are not matched. It should only keep the match. Now if we dont have any match, we try orthologs.

The `match_input_genes <- !all(input_genelistUP() %in% probes)` should do the job, but we need to test for every input and species

PR #829, opened 2023-12-21
Fix heatmap

ivokwee:
heatmap was not appearing for new calculated data. The test of '`genes$human_ortholog=="" `' was not sufficient because for Human '`genes$human_ortholog=NA`'.  This selection of gene_name, rownames(pgx$X) and human_ortholog, in relation with pgx$families and pgx$GMT.

Fix is temporary. This needs rethinking carefully before going to 'fully generalized features'

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,8 @@ data_shared/*
 trace/*
 logs/*.rds
 /logs/
+USER_INPUT/
+user_input/
 
 # Sensitive files #
 ####################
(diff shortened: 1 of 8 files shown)


mauromiguelm (changes requested):
needs refactor, I would apply same solution as https://github.com/bigomics/omicsplayground/pull/830

phisanti:
@mauromiguelm fixed some issue with non-human species. Ready to merge

PR #867, opened 2024-02-13
Fix cytoplot

ivokwee:
**problem:** Cytoplot in singlecell board does not render properly after last update

**cause:**  partial_bundle() in ICP plot is interfering with other plotly plots. Removed.

**learn**: be careful using partial_bundle() when having multiple plotly plots

Before:
image

After:
image


mauromiguelm (changes requested):
cytoplot is fine, but after reviewing snapshots, the board.signature has significant changes fold change in contrasts, so this needs to be reviewed before merging

PR #873, opened 2024-02-20
Fix #868: color plots that have just one level

### LINKED ISSUE #868: Colour in cell profiling plots no longer displaying correctly in v3.3.0-beta16-master240213
I noticed that after the update this week, the cell profiling plots are no longer displayed with the correct colours.

This is what I get using the same dataset

In v3.3.0-beta9-master231212:

image

In v3.3.0-beta16-master240213:

image

We lost the red colour along the way and the gray colour makes the plot indistinguishable from poor matches, which are a paler shade of grey and, bizzarely, red.

ESCRI11:
This closes #868 

## Description
Plots with just one level were not colored properly. Now they are


ESCRI11:
Changed colour defaults, so snapshot difference is OK

PR #890, opened 2024-03-08
Fix #877: move highlights to shiny modal

### LINKED ISSUE #877: Video does not stop playing when modal (or even opg session) is closed
video never stops (chrome latest)

image

ESCRI11:
This closes #877 

## Description
By moving to default `shiny` modals for the "BigOmics Highlights" we ensure that when dismissed it gets removed from the DOM, stopping the video playback. No need to use `bsutils` modals when having standard `shiny` ones.

Aesthetics changed a little bit. 

image


mauromiguelm (changes requested):
done:
- close buttons missing was added by bsutils header
- similar color to header added, but style of bsutils does not match shiny modal


to review:
- shiny modal takes longer to load content
- we need a background color to make the navigation bar visible, or use same styles to the navigation bar

PR #899, opened 2024-03-21
Update requirements and Dockerfile

mauromiguelm:
This pull request updates the requirements file and Dockerfile to include necessary package installations and dependencies. It also cleans up unnecessary installations and fixes a minor issue in requirements.R.


ivokwee:
OK. It seems to build correctly. I have added 'recount' to the requirements as playbase could not install without it.

PR #902, opened 2024-03-25
Invite popup

ivokwee:
Implements an 'invite!' button in the top menu bar that creates a popup to invite friends. An email is send using blastula to the friend and a copy to bigomics support. Invites get centrally logged into a file etc/INVITES.log.


mauromiguelm (approved):
- redirect to a new tab (done)
- redirect to a page that exists (done)

PR #931, opened 2024-05-06
Fix #922: 'Markers plot" under "Test geneset" gives an error message in v3.3.2-master240416

### LINKED ISSUE #922: 'Markers plot" under "Test geneset" gives an error message in v3.3.2-master240416
This error was reported by one of our user and I am getting the same error message with different datasets as well:

image

ESCRI11:
This closes #922 

## Description
On the base docker build, the `matrixStats` fixed version (https://github.com/bigomics/omicsplayground/blob/master/dev/requirements.R#L198) was overwritten when installing `playbase` (https://github.com/bigomics/omicsplayground/blob/master/docker/Dockerfile.base#L74). Having the `matrixStats` package with a newer version produces the bug.

This has been fixed by installing the correct version of `matrixStats` after `playbase`. 

Also, a temporary fix has been added to the regular `Dockerfile` to make sure next time we deploy the issue is gone. That temporary fix will be removed when a new base docker image is deployed.


mauromiguelm (approved):
I built a new docker and it works

image

PR #932, opened 2024-05-08
Fix #928: New feature request: colour patterns in the Volcano plots

### LINKED ISSUE #928: New feature request: colour patterns in the Volcano plots
This is a feature that has been frequently asked by users and I figured I should add it here as a request. Many users would like to be able to see Up- and down-regulated genes/proteins coloured red and blue respectively, as is frequently done in publications.

I think either changing the default colours to reflect that or alternatively adding a simple editing options to do so would be a popular feature with our users.

ESCRI11:
Next to https://github.com/bigomics/playbase/pull/108, this closes #928 

## Description
Added the requests by @mauromiguelm: Give user option to disable coloring by FC

### DIFF
--- a/components/board.enrichment/R/enrichment_plot_volcano.R
+++ b/components/board.enrichment/R/enrichment_plot_volcano.R
@@ -12,6 +12,17 @@ enrichment_plot_volcano_ui <- function(
     width) {
   ns <- shiny::NS(id)
 
+  plot_opts <- shiny::tagList(
+    withTooltip(
+      shiny::checkboxInput(
+        inputId = ns("color_up_down"),
+        label = "Color up/down regulated",
+        value = TRUE
+      ),
+      "Color up/down regulated features.",
+      placement = "left", options = list(container = "body")
+    )
+  )
 
   PlotModuleUI(
     ns("plot"),
@@ -23,6 +34,7 @@ enrichment_plot_volcano_ui <- function(
     caption = caption,
     plotlib = "plotly",
     plotlib2 = "plotly",
+    options = plot_opts,
     download.fmt = c("png", "pdf")
   )
 }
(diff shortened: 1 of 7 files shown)


mauromiguelm (changes requested):
1. in genesets > geneset expression, the volcano does not work when no gene is selected in table
image


2. genesets > geneset expression: No gene is highlighted when a gene is selected on table

image

3. genesets > geneset expression: color up/down does not work
image

3. Diffexpr> volcano by comparison (unrelated to the PR)  how difficult is to select the gene across all comparisons and highlight them in each volcano? It would be nice to have same behavior as DiffExpr
image

ESCRI11:
@mauromiguelm

ESCRI11:
1. This already happens on free/pro, will add a new issue to solve this as it is not related to this PR
image
2. and 3. I can't reproduce. Have you installed the latest playbase? Here is how it looks for me
image
image

4 I think it is not that difficult, however we have to also consider implementing into Diff Expression multiple volcano plots, should I open a PR?
image

mauromiguelm:
@

> 1. This already happens on free/pro, I will add a new issue to solve this as it is not related to this PR
>    image
> 2. and 3. I can't reproduce. Have you installed the latest playbase? Here is how it looks for me
>    image
>    image
> 
> 4 I think it is not that difficult, however we have to also consider implementing into Diff Expression multiple volcano plots, should I open a PR? image


> 1. This already happens on free/pro, I will add a new issue to solve this as it is not related to this PR
>    image
> 2. and 3. I can't reproduce. Have you installed the latest playbase? Here is how it looks for me
>    image
>    image
> 
> 4 I think it is not that difficult, however we have to also consider implementing into Diff Expression multiple volcano plots, should I open a PR? image

1. perfect, thats definitely a new issue.
2-3. yes, this PR branch is installed.. DiffExpr works fine, Gene expression not. Not sure if its related to dependencies
4. Its confirmed, maybe I have different dependencies... not sure how we can fix this besides building a docker... should I do it?

One more thing I saw is that the snapshot test failed, I think it used playbase@main instead or the right PR... very strange as it should match the two... any idea?

PR #957, opened 2024-06-03
fix: detect colored columns after modifying `df`

ESCRI11:
## Description
There was a small bug on differential expression analysis table. The columns to be color coded, were detected before modifying the `df`, therefore, there was no coloring applied.

By detecting the columns after modifying `df`, the colors are back.

### Before
image

### After
image


mauromiguelm (approved):
really nice, great work @ESCRI11!

PR #959, opened 2024-06-04
feat: plotly activation matrix

ESCRI11:
## Description
New activation matrix using plotly.

Goals are: improved readability, avoid overlapping/too crowded axis.

### Matrices changed
Pathway analysis: WikiPathWays, Reactome, GO Graph
SystemsBio: Drug connectivity

#### Small
image
#### Enlarged modal
image

#### Small
image
#### Enlarged modal
image


mauromiguelm (changes requested):
- in previous activation matrix, few cases were highlighted (see BG-STK3359 in drugconnectivity), now the same drug goes from red to blue, which is a different result as before. Which is correct?
- There is too much blue in the new plots.. maybe apply a color gradient that highlights extreme changes and not proximity to null (zero) effect?

ESCRI11:
Thank you for your comments Mauro! Agreed on everything; more important, good catch on the drugconnectivity plot, there was a mistake on the color axis.

I have changed the following.

1. Smoothed the color scales so it highlights better the extreme cases. (Figure 1)
2. (drugconnectivity) As per the old implementation, only use white to red color scale (also smoothed as per point 1) (Figure 2/3)
3. Inverted the Y-axis to match the old plots.

### Figure 1
image
### Figure 2
image
### Figure 3
image

Let me know what you think!

mauromiguelm (approved):

PR #964, opened 2024-06-11
Fix #962: allowing users to highlight specific genes on the Gene UMAP plot

### LINKED ISSUE #962: Request: allowing users to highlight specific genes on the Gene UMAP plot
This request came from a user at Emory University:

_“I want to mark my genes of interest on the UMAP. How can I pick and mark the genes on the UMAP?”_

Ivo suggested turning it into a Github issue, so here it is.

ESCRI11:
This closes #962 

## Description
This adds a `<custom>` option to filter out genes which opens a new conditional panel where the genes of interest can be inputed. Also, I have moved the `filteredGenes` function from the `featuremap_plot_table_gene_map.R` file to `featuremap_server.R` so that we can re-use it on the signature UMAP plot on the future if required.

image

### DIFF
--- a/components/board.featuremap/R/featuremap_plot_table_gene_map.R
+++ b/components/board.featuremap/R/featuremap_plot_table_gene_map.R
@@ -68,7 +68,7 @@ featuremap_plot_gene_map_server <- function(id,
                                             pgx,
                                             plotUMAP,
                                             sigvar,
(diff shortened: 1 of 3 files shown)


mauromiguelm (approved):
works really well, no changes needed

Issue #955, opened 2024-05-30
Data upload: Large samples.csv file not correctly showing


zitoa:
Data upload: Large samples.csv file not correctly showing

PR #970, opened 2024-06-19
fix: use `req` with a logic condition

ESCRI11:
This closes https://github.com/bigomics/omicsplayground/issues/955

Use a conditional statement inside the `req` to avoid strange behaviour with `matrix`, `NA`s, etc.


ESCRI11:
This closes #955

ivokwee:
OK. checked. previously a character matrix that included NA would not show up. Now does.

PR #993, opened 2024-07-12
Feature: compare datasets compute button

ESCRI11:
Feature from client feedback.

## Description
On the compare datasets, placed a 'compute' button to redraw the plots instead of acting upon new contrasts added; this created a pain-point when users wanted to compare multiple contrasts of big datasets.


ivokwee:
I think you do not need to pass the compute button explicitly to all functions. the reactiveVal for input.button1 and input.button2 can function sufficiently as trigger (triggered on button). 

Also I do not like the duplication of code for plot_data and RENDER. I show in compare_plot_compare1_server how and extra function call_createPlot() can be called by plot_data and RENDER. 

Please refactor the rest.

PR #995, opened 2024-07-22
feat: check logarithm and show confirmation dialog at counts upload

ivokwee:
This adds a confirmaton dialog to convert log-values to intensities at user upload. Previously we detected and converted log values automatically and this was not always correct for examples with TMT fraction data. This PR adds a confirmation dialog before undoing the log. This works with the new playbase::is_logged() function that improves log  detection.


ivokwee:
Also files with name "expression.csv" are not automatically converted anymore. Expression.csv will be treated the same as counts.csv. This is also because in proteomics we will use expression.csv/abundance.csv  for linear intensities.

ivokwee:
I added removing negative values (in linear space). Negative values are set to zero.

PR #1010, opened 2024-08-06
feat: updated user details table

ESCRI11:
## Description
Small rework of the User profile table, which at this moment was pretty much out of service:

image

I have placed on it more relevant information for the user

image


ivokwee:
This needs more rethink and the 'upgrade' button is too early.

PR #1018, opened 2024-08-09
Fix computation spinner displaying when a dataset is computed in the background

mauromiguelm:
This pull request fixes the computation spinner.


ivokwee:
OK. working again. Merging.
Before:
image

After:
image

PR #1019, opened 2024-08-09
Metabolomics in Omics Playground

mauromiguelm:
# Omics Playground

**In OmicsPlayground**, this pull request adds support for the metabolomics data type. It includes changes to the UI to allow the selection of the probe type for metabolomics data. It also updates the translation file to include translations for the metabolomics data type. Additionally, there are changes to the server logic to handle the detection of the probe type for metabolomics data.

# Playbase

- Update the logic to accept metabolic genesets via custom genesets.
- Bring back code that add "test" GMTs in case no mapping is done between annotated metabolites and GMT file. This will allow generalized features in the future.
- Add a new annotation for metabolomics
- Add tests to make sure annotation is performing correctly.

# Playdata

- Add metabolic pathways in gmt format in `playdata::REACTOME_METABOLITES`
- Add an annotation table supporting several of the major ids (HMDB, PubChem, KEGG...) in  `playdata::METABOLITE_ANNOTATION` covering ~250'000 ChEBI metabolites as main ID
- Add metadata information (metabolite name...) for each ~250'000 ChEBI metabolites as in `playdata::METABOLITE_METADATA`

**Requires:**

**playbase:** https://github.com/bigomics/playbase/pull/131
**playdata:** https://github.com/bigomics/playdata/pull/9


ivokwee:
great but let's wait until beta gets more stable.

PR #1028, opened 2024-08-13
feat: probe filter defaults for proteomics

ESCRI11:
## Description
All probe filter are set as disabled by default for proteomics data.

image

### DIFF
--- a/components/app/R/global.R
+++ b/components/app/R/global.R
@@ -186,6 +186,14 @@ opt.file <- file.path(ETC, "OPTIONS")
 if (!file.exists(opt.file)) stop("FATAL ERROR: cannot find OPTIONS file")
 opt <- playbase::pgx.readOptions(file = opt.file, default = opt.default) ## global!
 
+message("\n************************************************")
+message("************* SETTING DEFAULTS ***************")
+message("************************************************")
+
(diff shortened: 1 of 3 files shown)


ivokwee:
Can we see if it is possible to have user/deployment defined defaults in a file like an DEFAULTS or defaults.yaml file? In that case we can set for any deployment different setting. No need to do it on User level, just Deployment level like the current global OPTIONS file.

ESCRI11:
@ivokwee Sounds like a good idea, working on it

ESCRI11:
@ivokwee Check the last 2 commits for the first iteration of your idea. If looks good to you I can extend it to different defaults. Let me know

ivokwee:
Nice. But I took out the datatype dependency (if proteomics). I think this is more a client preference. and to keep it simple. perhaps in the future we need it but I want to keep any datatype switches at a minimum.

ivokwee:
We can extend this to many other places if needed. Not sure if something "automatic" by writing to Shiny's input could be done. For now this is minimal and good.

ivokwee:
Rembmer. For MPOC we need to modify the DEFAULTS file at deployment.

PR #1029, opened 2024-08-14
Add: checking of organism and datatype in ExtendedTask

ivokwee:
Adding checking and warning of wrong organism (cross checking only human, mouse, rat) in ExtendedTask. Only check the 3 major organisms, this is reasonably fast because we have orgDb for them installed as packages.

image

Is also checks if the probetype is proteomics but the selected datatype is different than 'proteomics'

image

### DIFF
--- a/components/board.upload/R/upload_module_computepgx.R
+++ b/components/board.upload/R/upload_module_computepgx.R
@@ -596,11 +596,6 @@ upload_module_computepgx_server <- function(
           sendSuccessMessageToUser = sendSuccessMessageToUser
         )
 
-        ## Test check dim(counts) & dim(X)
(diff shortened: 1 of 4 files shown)


ivokwee:
Improvement could be to change or ask the user for changes in organism or datatype in the popup.

mauromiguelm (approved):
organism:
- No need to detect species as user already selected species; prone to error
- Since user cannot change organism without closing wizard (as organism is not in pop-up), a new pop-up would make sense

organism not tested in orgdb:
- If organism is not among the three main species, we could use ah/orthogene to validate. Not having check will lead to users uploading wrong species/probes.

Other than that code works fine for me, so approved

ivokwee:
the code checks (as before) the user specified organism AND the 3 other species. This is just an extra check whether the user was 'lazy' and did not selected the organism correctly. Like having mouse data, but leaving the organism to human.

PR #1031, opened 2024-08-14
fix issue with flexbox container overflow in upload wizard

mauromiguelm:
**Reason:**

While fillRow is based on flexbox, it is based on bootstrap 3 which caused a conflict and incorrect fillign behavior:

**problem:**
image

**old code snippet from tree:**
image

Replacing to the new bootstrap 5 the fill item is correctly assigned to the container item and it fills the space properly.

**solution:**
image

**code after fix:**
image


ivokwee:
Nice

PR #1045, opened 2024-08-20
fix: color ALL significative genes

ESCRI11:
## Description
Not converting the thresholds to class numeric was causing some detection issues

### Before (example-data)
image

### After (example-data)
image


ivokwee:
wow. I would not have expected such behavior... what type was is then? weirdness of R...

PR #1059, opened 2024-08-27
Fix issues with subsetting and NA values in gene sets affecting features not mapped to symbol

mauromiguelm:
This pull request addresses several issues related to subsetting and NA values in gene sets. NAs in the subset of rnaX caused the entire enrichment module to crash.


ESCRI11:
Tested - works good with metabolics + other datasets I have

ESCRI11 (approved):

PR #1061, opened 2024-08-27
feat: metabolite defaults for testgeneset

ESCRI11:
## Description
The defaults for GeneSet/Test geneset were gene names. Updated the defaults when metabolomics dataset is loaded to reflect ChEBI ID's. Also added some default examples: glycolisis, citric acid cycle and urea cycle.

image


ESCRI11:
@mauromiguelm Can you test this with a non-human dataset?

mauromiguelm (changes requested):
great improvement!
feedback:
- some metabolites are not being colored when significant
image

- enrichment has an error
image
- I dont know what this plot does, but weird
image

ESCRI11:
@mauromiguelm Thank you for the feedback, the plot has been fixed on commit [801c367](https://github.com/bigomics/omicsplayground/pull/1061/commits/801c3674b054cbb9fb177fed5995be7bc5cb5e5d).

Regarding the last issue you point out, I have taken a look and I am guessing it has to do with not a single "geneset" being significant, since this pvalue correction https://github.com/bigomics/omicsplayground/blob/master/components/board.signature/R/signature_server.R#L293 makes them all 1, before they are 0.7/0.8 -- no significative values at all.

I am not sure how to test my hypothesis.

PR #1063, opened 2024-08-28
feat: updated plotmodule cards

ESCRI11:
## Description
Make use of different `bslib` functions to build the plot card. This ensures that the plot displayed is properly height adjusted when using cards; before that was not the case, the height had to be statically set, which led to problems with different screen sizes, now the plot automatically uses the whole card space.

### Before
image

### After
image


ESCRI11:
Wait for merge, found an issue with the download button

ESCRI11:
Ok now it is fixed

mauromiguelm (approved):
really well spotted, it also fixes the missing "labels" for features that are not significant.

PR #1066, opened 2024-08-29
Feat: download tables as excel or csv

### LINKED ISSUE #1065: Download tables as `*.xlsx`
From client feedback: Would be useful to be able to download tables directly as `*.xlsx` instead of `*.csv`.

ESCRI11:
This closes #1065 

From client feedback: Would be useful to be able to download tables directly as `*.xlsx `instead of `*.csv`. 

For that, I have added a switch to select the format to download.

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -40,6 +40,7 @@ Imports:
     magrittr,
     markdown,
     miniUI,
+    openxlsx,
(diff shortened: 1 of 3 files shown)


ESCRI11:
Also added the option to the Plot Module 
image

Issue #1014, opened 2024-08-08
Error in updateDatasetFolder: attempt to set 'colnames' on an object with less than two dimensions


mauromiguelm:
dataset: 730e1c337

This is a critical system maintained by @ivokwee so I am tagging you Ivo.

```r
[updateDatasetFolder] missing sigdb. Creating new sigdb file.
WARNING:: SD is constant. Skipping SD reduction...
calculating PCA 2D/3D...
Error in `colnames<-`(`*tmp*`, value = paste0("PC-", c("x", "y"))) : 
  attempt to set 'colnames' on an object with less than two dimensions
Calls: <Anonymous> ... pgx.clusterBigMatrix -> pgx.clusterMatrix -> colnames<-
```

ivokwee:
dataset 730e1c337 computes successfully on Omics Playground v3.5.0-beta10+fix-1014240906

updateDatasetFolder finishes succesfully:
![image](https://github.com/user-attachments/assets/449dc81c-2491-4b90-b3ef-ee4396d568a1)


I suspect probable cause in clusterMatrix is matrix is constant or has NA. Perhaps when zero probes are not filtered out? I will test more.

ivokwee:
I have tested pgx.clusterMatrix with matrix completely zero. Same "SD constant warning" but could not replicate crash in colnames. Closing this issue.

![image](https://github.com/user-attachments/assets/3aa0104f-773f-4b2a-a21b-1c33c4f6d69f)

PR #1080, opened 2024-09-06
fix detect_task error #1014

ivokwee:
fixes issue #1014 . return value of detect task is not NULL but NA. Adding more error detection statement.


ivokwee:
Sorry. Upload of params.RData was solved above. Actual bug in updateDatasetFolder still to be confirmed and solved.

Issue #1081, opened 2024-09-06
`Settings` cog container dissapears if click happens *inside* the container; should only dissapear by outside click


mauromiguelm:
` stopPropagation` if click is inside container is missing somewhere :
https://github.com/user-attachments/assets/14481a75-b693-48f6-9df6-74b8799c3ee8

PR #1082, opened 2024-09-06
Update labeltype logic and UI controls

mauromiguelm:
requires fix https://github.com/bigomics/omicsplayground/issues/1081

Genomics datasets can be visualized with symbol.

Metabolomics datasets have to be visualized with metabolites names, as all symbols are numeric.

This PR implements a global label change between feature, symbol and name.

- Adding a global labeltype variable that is initialized when PGX is loaded.

- Applying the same solution to the clustering board.

- Fixing an issue where the playbase::ggVolcano plot was not showing correct labels for metabolomics.

- Moving the labeltype logic to trigger at every dataset change.

- Removing the labeltype selector for the clustering splitmap globally.

- Adding a global UI control in the settings kog wheel.

These changes improve the flexibility and functionality of the labeltype feature in the codebase for different data type requirements.


ivokwee:
Not sure about this. I understand to try to centralize labeltype a bit because it is used in many plots, but putting it in the cog wheel I find the selection is too "far" from the plots. Trying to find the option in the cog wheel seems to "far". Also it seems not consistent because some plots (like 'Top correlation' barplot in DataView) do not change when selecting global labeltype.

mauromiguelm:
> Not sure about this. I understand to try to centralize labeltype a bit because it is used in many plots, but putting it in the cog wheel I find the selection is too "far" from the plots. Trying to find the option in the cog wheel seems to "far". Also it seems not consistent because some plots (like 'Top correlation' barplot in DataView) do not change when selecting global labeltype.

one option is to place in all or all relevant sidebar settings, but it will take space and some boards (clustering) is already cluterred with settings..

I find it matches nice with the plot legends, but open to ideas

mauromiguelm:
**top correlation:** I tried but the labels become too big. You can hover over and see the gene title.

**gene selector:** I now updated the gene selector, so you can search for metabolite names.

image

ESCRI11 (approved):

PR #1087, opened 2024-09-09
fix grey dots due to incomplete selection

ivokwee:
fix grey dots in volcano due to incomplete selection when multiple features map to same symbols.

Before:
image

After (using PTM data):
image


ivokwee:
Note: I guess lots of confusion because naming of variables do not reflect whether we mean gene symbol, feature, or gene title. we (still) use sel.genes if now it should be sel.features, 'names' is ambigious to mean gene name or gene title. Would be good to clean up.

PR #1092, opened 2024-09-10
Fix #1083: Nodes in GO graph cannot be moved anymore

### LINKED ISSUE #1083: Nodes in GO graph cannot be moved anymore
Nodes in GO graph cannot be moved anymore (v3.5.0-beta10). Previously we could drag the nodes to change the layout, specially in case the node labels overlap. Dragging seems not working anymore. Tested both small and big plot.

ESCRI11:
This closes #1083 

## Description
It seems that for visnetwork to properly work, it has to be somehow initialized before loading the data+modules. Probably it has to do with some JS initialization not happening.

A little bit confusing solution nevertheless.

### DIFF
--- a/components/app/R/ui.R
+++ b/components/app/R/ui.R
@@ -37,7 +37,6 @@ app_ui <- function(x) {
     #-------------------------------------------------------
     ## Build USERMENU
     #-------------------------------------------------------
-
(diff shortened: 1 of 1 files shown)


ivokwee:
OK. works.

Issue #1052, opened 2024-08-26
Similar experiments: cache results and offline compute


mauromiguelm:
similar experiments sometimes contain many datasets (GEO and certain clients), which causes the computation to take minutes.

We should try to compute these offline, and if possible cache the results, as switching between the same databases causes recomputation multiple times (no caching).

PR #1097, opened 2024-09-12
save connectivity results of USER sigdb to avoid recomputation

ivokwee:
This addresses issue #1052 and saves the results of the USER sigdb connectivity computation into the pgx object to avoid computation each time. 

Code should also detect if computation object should be recomputed if USER folder has new files. Please note: saving the connectivity results into the pgx will alter the original file and therefore the time stamp.

Needs latest updated playbase with modified pgx.load() and pgx.save(). They record the original filename in the pgx object as pgx$filename so the pgx can be saved to the original file location.

### DIFF
--- a/components/board.connectivity/R/connectivity_server.R
+++ b/components/board.connectivity/R/connectivity_server.R
@@ -227,6 +227,9 @@ ConnectivityBoard <- function(
           check.sigdb = TRUE,
           verbose = FALSE
         )
+        
+        dbg("[compute_connectivity] need_update = ", need_update)
+        dbg("[compute_connectivity] names(pgx$connectivity) = ", names(pgx$connectivity))        
 
         if (need_update || !file.exists(sigdb.file)) {
           pgx.showSmallModal("Updating your signature database<br>Please wait...")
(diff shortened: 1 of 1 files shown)


mauromiguelm (approved):

mauromiguelm (changes requested):
at least in my computer, need_update always returns true

```r
need_update <- playbase::pgxinfo.needUpdate(
        pgxdir,
        check.sigdb = TRUE,
        verbose = FALSE
)
```

which causes re-computation when the board is re-loaded for the same dataset

```r
Browse[1]> need_update
[1] TRUE
Browse[1]> !file.exists(sigdb.file)
[1] FALSE
Browse[1]> need_update || !file.exists(sigdb.file)
[1] TRUE
```

ivokwee:
What do you mean with "invalid pgx", some slots missing because of incomplete calculation?

ivokwee:
Mine 'needUpdate' give false but I think all my pgx are 'valid'
image

mauromiguelm:
> What do you mean with "invalid pgx", some slots missing because of incomplete calculation?

the ones that do not pass https://github.com/bigomics/playbase/blob/f8acc95e6b1374d21ad3d9ad45ee76e386450d03/R/pgx-init.R#L48-L54
We could have a check before incorporating pgx into datasetInfo folder, so the ones not passing tests above will not be included and not trigger recomputation.. but thats an edge case probably

ivokwee:
OK. merging. Invalid PGX should not happen. this is a separate problem.

PR #1098, opened 2024-09-13
fix: add i18n to dev boards

ESCRI11:
## Description
Fix for the snapshots so we can use them again. i18n was missing on the dev boards for it to work again.


mauromiguelm (approved):
Good but we have quite a few snaps with problems, some examples
dataview_Counts table
intersection_Pairwise scatter

connectivity_FC correlation

ESCRI11:
> Good but we have quite a few snaps with problems, some examples

Great, now time to work on them and move forward with testing!

PR #1106, opened 2024-09-20
Fix #1105: return to defaults if not pairwise

### LINKED ISSUE #1105: pairwise methods get stuck after removing 1vs1 comparisons
When we create a new pairwise comparison, if we create one with only 1 sample and then remove it, under the computation options I can still only see limited options for pairwise comparisons (i.e. only ttests) and GSEA. It should reset back to all the options.

ESCRI11:
This closes #1105 

## Description
By setting the `else` on the update function, we make sure to go back to defaults when all contrasts have more than 1 vs 1 comparissons.

### DIFF
--- a/components/board.upload/R/upload_module_computepgx.R
+++ b/components/board.upload/R/upload_module_computepgx.R
@@ -59,8 +59,19 @@ upload_module_computepgx_server <- function(
         }
       )
 
-      ## GENETEST.SELECTED <- c("trend.limma", "voom.limma", "deseq2.wald", "edger.qlf")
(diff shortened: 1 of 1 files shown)


ESCRI11:
Also following Alex suggestion, I added different defaults for gsets when it comes to proteomics/other data. (https://github.com/bigomics/omicsplayground/pull/1106/commits/7609bef3e45943340eca7e53137b817257e98007)

"it seems now even for RNAseq data the default is to use proteome-centric options (Ttests and limma)."

mauromiguelm (approved):
works perfectly!

PR #1122, opened 2024-09-27
Refactor PCSF board

ivokwee:
Refactoring PCSF board.

1. Now using new computePCSF() and plotPCSF() functions from playbase
2. Using 1, is more consistent with report
3. Using 1, moves more data prep and plot prep code to playbase (which is a-good-thing-to-do..)
4. Add initial network size S, M, XL. This will indirectly increase solution size, like beta but not the same.


ivokwee:
Note: UI has changes. This will of course raise error in ShinyTest.

PR #1123, opened 2024-09-30
Fix #1118: Subset selection does not work in feature UMAP (genes)

### LINKED ISSUE #1118: Subset selection does not work in feature UMAP (genes)
In the Clustering > Features > Gene > Feature UMAP, the Box select does not seem to work (it does not subset the Feature table below). At the Geneset level the subsetting of the table works.

[email cg@tulane,  Wed, Sep 25,]
[IK. 26/9 confirmed,]

ESCRI11:
This closes #1118 

## Description
We were not grabbing the selected genes on UMAP plot. With this fix we detect the brush and subset the table accordingly -- pretty much copy and paste from geneset tab where this is already implemented.


ivokwee:
strange. because i remember it worked before.

PR #1159, opened 2024-10-08
Fix #1158: Test genesets: 'x' must be an array of at least two dimensions

### LINKED ISSUE #1158: Test genesets: 'x' must be an array of at least two dimensions
Hubspot ticket 9347794935

image

ESCRI11:
This closes #1158 

## Description
as always, single case breaks things. adding `drop=FALSE`.

image


ESCRI11:
also added a small change to the handling of repeated errors; we now check just the error message not the whole traceback -- should be more effective

PR #1183, opened 2024-10-15
Fix #1182: Feature set ranking crash for restrictive sample filter: undefined columns selected

### LINKED ISSUE #1182: Feature set ranking crash for restrictive sample filter: undefined columns selected
Hubspot ticket 10433763061

example-data

image

ESCRI11:
This closes #1182

Added validator for cases when filtering is too restrictive.

image


ESCRI11:
also added `req` to solve ticket 137523500

PR #1215, opened 2024-10-21
Fix #1214: Remove outliers not working as expected

### LINKED ISSUE #1214: Remove outliers not working as expected
The outliers highlighted on the plot are not removed / other samples are removed.

From FP.

ESCRI11:
This closes #1214 

## Description
This subseted matrix is the same matrix used to display the outliers plot, otherwise there is a mismatch between outliers on the plot and outliers at this stage. By doing the same subset we remove the outliers marked on the plot (expected behaviour).

This PR can be validated using example upload data and setting remove outliers threshold to 3. Before, no sample was removed, after this PR the correct sample will be removed.


zitoa:
playbase::detectOutlierSamples selects the top 4000 by sd.  maybe we leave that as default and plot using 4000? Ok unless is much slower ...

ESCRI11:
Sounds good to me

PR #1223, opened 2024-10-22
Fix #1222: Correlation analysis: unavailable custom gene crahes plots

### LINKED ISSUE #1222: Correlation analysis: unavailable custom gene crahes plots
Reproducible on example-data + hubspot ticket 15931081941 (and others)

If the genes inputed as <custom> are not present on the data, the plots crash.

image

Add a shiny validator to handle this case.

ESCRI11:
This closes #1222

## Description
If inputed <custom> genes are not found (any of them), return a validator message, otherwise the plots crash.

image


zitoa (approved):
Approved

PR #1224, opened 2024-10-22
Add WARNING message about DGE testing in presence of NAs

zitoa:
In presence of NAs, inform the user which DGE tests can be performed and which one can't.  Advice user to review their choices.


ivokwee:
Probably good to "grey out" the methods that are not applicable.

ivokwee:
Probably good to "grey out" the methods that are not applicable.

ivokwee:
Please do not commit VERSION and styles.min.css files

PR #1275, opened 2024-11-20
Fix #1274: Sample QC: subscript out of bounds

### LINKED ISSUE #1274: Sample QC: subscript out of bounds
Hubspot ticket 33474328783

image

ESCRI11:
This closes #1274 

This issue comes from the group by having empty group "". All is OK but subsetting a data frame using column names fails, therefore we can match the indexes and properly display the grouping.

image


ESCRI11:
Also added a quick fix where we were using `Y` instead of `samples` for Hubspot ticket 31918942413

PR #1363, opened 2025-04-02
rm ia contrasts fr menu and add in table

zitoa:


zitoa:
I see IK reviewed this. I am merging it @ESCRI11

PR #1368, opened 2025-04-08
Remove computed interaction contrasts from all modules (except DataView > Contrasts table; TimeSeries; Differential expression).

zitoa:
Remove computed interaction contrasts from all modules (except DataView > Contrasts table; TimeSeries; Differential expression).

### DIFF
--- a/components/board.pathway/R/pathway_plot_go_actmap.R
+++ b/components/board.pathway/R/pathway_plot_go_actmap.R
@@ -74,7 +74,6 @@ functional_plot_go_actmap_server <- function(id,
     id, function(input, output, session) {
       shiny::observe({
         shiny::req(pgx$X)
-        #ct <- colnames(pgx$model.parameters$contr.matrix)
         ct <- playbase::pgx.getContrasts(pgx)
         ct <- sort(ct[!grepl("^IA:", ct)])
         selected_ct <- head(ct, 8)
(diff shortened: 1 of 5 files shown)


components/board.pathway/R/pathway_plot_go_actmap.R
```diff
@@ -88,11 +89,10 @@ functional_plot_go_actmap_server <- function(id,
       plot_data <- shiny::reactive({
         shiny::req(pgx$meta.go)
         score <- pgx$meta.go$pathscore
+        score <- score[, input$selected_contrasts, drop = FALSE]
```
ESCRI11, line 92: Why is this needed?

ESCRI11 (changes requested):
Maybe instead of

```
      ct <- playbase::pgx.getContrasts(pgx)
      ct <- sort(ct[!grepl("^IA:", ct)])
```

Maybe, we can modify `playbase::pgx.getContrasts(pgx)` to have a new argument (default to `FALSE`) to just call `playbase::pgx.getContrasts(pgx, remove_ia = TRUE)`. For cleaner code.

Also just remove the old ways (e.g. `      #ct <- colnames(pgx$model.parameters$contr.matrix)`) instead of commenting them.

components/board.pathway/R/pathway_plot_go_actmap.R
zitoa: Not needed i think

zitoa:
@ESCRI11 Cleaning done. If you're ok we can just merge and move on

ESCRI11 (approved):

PR #1375, opened 2025-04-15
scRNA-seq: Test signatures > Markers

zitoa:
Test signatures > Markers. Usually very slow for single cell due to gsva use and shiny single samples reactives. Introduced supercell there with low gamma (strong compression)- speeds up computation a lot.


ESCRI11 (approved):

zitoa:
Test data for the others to replicate and review: any scRNA-seq data (no metacell) applies

PR #1376, opened 2025-04-16
Use symbol on gsea comp

ESCRI11:
On species other than human/mouse (found issue on arabidopsis and dmelanogaster-flybase), we should use symbol on `rnk0`, otherwise, there is a missmatch.

image

After fix:

image


zitoa:
@ESCRI11 I think we should merge this into devel rather than master?

ESCRI11:
@zitoa True! thank you, just changed it to devel

zitoa (approved):

PR #1378, opened 2025-04-16
For datasets with many contrasts: limit contrasts displayed in volcanos for better visual

zitoa:
GeneSets > Geneset enrichment > Volcano by comparison > Volcano plots for all contrasts.

In case of many contrasts, (typically for scRNAseq but also large bulk data), the Volcanos are very messy. Furthermore, the "Contrast:" selected in the drop-down menu on the right pulled the genesets for that contrast in the table, but may not be very visible in Volcano plot when there are many contrasts. Here 2 changes:
1.  Added plot option to allow user to choose how many contrasts to visualize;
2.  Ensured that the selected contrast in the drop-down menu is always displayed.

I also removed the previous plot option "enrch_volcanoall_subs" which was unused.

Visual is now clean as per default.

### DIFF
--- a/components/board.enrichment/R/enrichment_plot_volcanoall.R
+++ b/components/board.enrichment/R/enrichment_plot_volcanoall.R
@@ -21,14 +21,9 @@ enrichment_plot_volcanoall_ui <- function(
       "Scale the volcano plots individually per method..",
       placement = "right", options = list(container = "body")
     ),
-    #withTooltip(
-    #  shiny::radioButtons(ns("enrch_volcanoall_subs"), "Subsample plot",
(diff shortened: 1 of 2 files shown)


ESCRI11 (changes requested):
looks good to me; please remove the commented lines + default visualization values should be able to display all contrasts on example-data, so maybe I would change 4 to 6

zitoa:
@ESCRI11 done

ESCRI11 (approved):

ivokwee:
@zitoa For the PR's you should always include proof of before and after, like a screenshot or terminal output. Also describe and if possible provide the test data for the others to replicate and review.

PR #1380, opened 2025-04-22
Devel close message

ESCRI11:
From office conversation:

- Display a warning message when computation is in progress and user closes tab/browser.
- Make sure this only happens while data is *actually* being computed.

Before: (compute in progress)
[Screencast from 2025-04-22 17-42-29.webm](https://github.com/user-attachments/assets/61eb5485-4ada-4e34-86cb-854f454da8c8)

After: (no compute in progress)
[Screencast from 2025-04-22 17-32-25.webm](https://github.com/user-attachments/assets/b65a3d0c-eee8-41fb-be99-b37012fab3ce)

After: (compute in progress)
[Screencast from 2025-04-22 17-40-40.webm](https://github.com/user-attachments/assets/73b5b977-4dd2-4e28-b1bc-48c6dffb9105)

Regarding the contents of the message: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

> Only show a generic browser-specified string in the displayed dialog. This cannot be controlled by the webpage code.

### DIFF
--- a/components/app/R/ui.R
+++ b/components/app/R/ui.R
@@ -68,6 +68,8 @@ app_ui <- function(x) {
         ##    gtag2, ## Google Tag Manager???
         shiny::tags$head(shiny::tags$script(src = "custom/temp.js")),
         shiny::tags$head(shiny::tags$script(src = "static/copy-info-helper.js")),
+        shiny::tags$script(src = "custom/close-message.js"
+  ),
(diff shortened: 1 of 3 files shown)


ivokwee:
OK. works. Custom message would be great but I see they disabled it on newer browsers.

image

PR #1392, opened 2025-04-28
Scatterplot pairs: use pairwise to compute `r`

ESCRI11:
Some datasets (example: salmon-ENSOKI from opg-exampledata) have some NaNs on the matrix used to compute `r` on the scatterplot pairs.

Use `pairwise` correlation to not get NA correlation.

### Before
salmon-ENSOKI_intersection_Pairwise_scatter

### After
image


zitoa:
That would be very similar to use impX.

zitoa (approved):

Issue #1372, opened 2025-04-10
problem with several table and plots


Ax-h3ll:
Hello,

I was using the plateform today and I had several problems with the display of table and plots:
I was mainly in the Expression menu.
in the differential expression menu:
- in the Fold change by comparison plot, not all contrasts are displayed.
- in the Gene sets with gene table, nothing is displayed, even if I select a gene.
Then also in the correlation analysis menu:
- in the Correlation scatter plots nothing was displayed.

Then the website crashed and now I am not able to log in. 
I can see the log in page and but when I enter my informations, the website keeps turning and turning but does not log on.

ivokwee:
Does this problem still exists?

Ax-h3ll:
yes!
I still do not have anything showing up in the geneset with gene table when I select a gene in the DE analysis section.
same with the fold change by comparision, does not show all contrasts from my dataset.

<img width="1157" alt="Image" src="https://github.com/user-attachments/assets/de1aa65c-822b-4ce0-933c-ba0f38ff3f3c" />

ivokwee:
OK. Thanks. I could reproduce the geneset problem (nr2) with some other mouse dataset. It seems that for non-human organism the selection does not react. I am going to look at this problem.

For the other problems, it would be great if you could make a screenshot and possibly share the data to reproduce the error. If not, it is sometimes difficult for us to reproduce and check the problem. Thanks.

ivokwee:
About first issue. "in the Fold change by comparison plot, not all contrasts are displayed.". Is that the "Volcano by comparison"? I see no issue. Perhaps when you more than 8-10 comparisons, it will not show some. We should implement some chooses in that case. Can you confirm you have many comparisons? Please resubmit this issue with example data, or pointer to dataset if still exists. 

Closing this multipel issue ticket. Next time please submit issues separated by issue.

ivokwee:
Geneset problem will be solved in PR https://github.com/bigomics/omicsplayground/pull/1397

ivokwee:
correlation graph fixed in PR #1402

PR #1397, opened 2025-04-29
fix gene-in-gsets table (sub-issue 1372)

ivokwee:
Sub-issue of issue #1372 :  Geneset table not showing genesets

Issue: code is still matching on human ortholog (perhaps old way). In newer pgx, seneset are by organism's native symbols.


ivokwee:
Confirmed. Screenshot before fix:
image

ivokwee:
Will be fixed in master (need to upmerge to devel). 

Solution:  match best column (symbol or ortholog) to GMT matrix. Older pgx may have GMT as human ortholog, newer pgx should use native symbol

image

ESCRI11 (approved):

PR #1404, opened 2025-04-30
Fix #1384: wrong gene/feature higlighted in MA plot

### LINKED ISSUE #1384: wrong gene/feature higlighted in MA plot
Bug. Wrong feature (here: A2M) is highlighted when gene/feature (here: TNIP3) is selected. Using example-data.

Image

ESCRI11:
This closes #1384

During devel of 4.0 labeltype switch at some point got unified using pgx$genes$gene_name(https://github.com/bigomics/omicsplayground/commit/bf913bd33eaf3c86519351c1dbe55b230735c38c)

However, here, we need to make sure that `label.names` is ordered the same as `res (and has same length):

image


ivokwee:
approved. thanks.

PR #1406, opened 2025-04-30
Fix #1405: Diff expression: title not translated to labeltype

### LINKED ISSUE #1405: Diff expression: title not translated to labeltype
Dataset: salmon-ENSOKI from opg-exampledata

Diff expression plot displays feature name, not symbol as selected

Image

ESCRI11:
This closes #1405 

Send correct labeltype translated (pgx$genes$gene_name) as the plot title argument

image


ivokwee:
Looks good. Using pgx$genes$gene_name is the preferred way now.

PR #1433, opened 2025-05-21
feat: label types must be diverse

ESCRI11:
Apply criteria to offer as "label type" variables that have at least 10% unique values (compared to the total number of features). Less than that is probably useless as a label for plots


zitoa:
Compare Signatures >> Pairwise scatter: change in labeltype not working

zitoa (approved):

PR #1437, opened 2025-05-23
Add CV Boxplot in Dataview >> SampleQC

zitoa:
Added coeff. of variation Boxplot in Dataview >> SampleQC.

Changes in Dataview >> SampleQC.
1. Barplot of overall proportion of major protein classes (originally at bottom left) is now optionally displayed when the option " Show overall proportions" in Proportion of major protein classes per sample/group barplot is checked (the two barplots have complementary informations).
2. After change in 1., a space was created at bottom right that i used for CV boxplot.

IK ok with that

Before (current)
image

PR:
Screenshot from 2025-05-23 15-06-25
Screenshot from 2025-05-23 15-06-33
Screenshot from 2025-05-23 15-06-41


ESCRI11 (changes requested):
I like it. Maybe remove the old code instead of commenting it, otherwise lgtm

zitoa:
Depending on how big is the PR, I'd leave it for a week or so usually. Regardless, I will remove from there and from any new one and merge.

PR #1439, opened 2025-05-27
fix: drop=false...

ESCRI11:
single contrast data fails


ivokwee (approved):
The infamous drop problem...

PR #1456, opened 2025-06-20
Fix/improve handling duplicated features

zitoa:
Addex box to enable selection of treatment of duplicated features by user: Average or Keep all. If "Average", duplicates (if any) are averaged as we did previously. If "Keep all", duplicated features (if any) are made uniques. If "Keep all", this fixes visualization of all features in Data View >> table, both in linear and in the log-space. A separate PR to rename linear to "raw" will be created. Another fix will be needed for Volcano plot and will require a separate PR. 

Screenshot from 2025-06-21 00-03-07

This is a sensitive PR. I will also review again it tomorrow/Monday and add changes if needed prior to merge.

### DIFF
--- a/bin/pgxcreate_op.R
+++ b/bin/pgxcreate_op.R
@@ -34,18 +34,19 @@ pgx <- playbase::pgx.createPGX(
   probe_type = params$probe_type,
   description = params$description,
   creator = params$creator,
-  batch.correct = params$batch.correct,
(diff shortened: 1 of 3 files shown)


zitoa:
Meeting summary:
1. Move to computation options
2. Default "Keep all"
3. Move code into pgx.createPGX.

ESCRI11 (approved):

PR #1461, opened 2025-06-25
fix: declar lato for pdf render on ggplot

ESCRI11:
Some ggplots (volcano for example) use lato font to match the platform.

However, when plotting it to pdf, it complains that the font is not available, so we have to manually add it for pdf to work fine.


ivokwee (approved):

ESCRI11:
no its fine, it grabs it from google fonts repo

PR #1495, opened 2025-07-11
Fix filter matrix for non-human organism in clustering server

ivokwee:
Fix filter matrix for non-human organism in clustering server. 
- Old code still assumed symbols on rownames. Need to match with pgx$genes.
- pgx$families are not consistently organism symbol or ortholog. This is well handled using playbase::map_genes(). Still it would be good to harmonize pgx$families using organism symbol for consistency.

NOTE: getFilteredMatrix.SAVE still needs to be removed if fix works.


ESCRI11 (commented):
Why do we have changes on MOFA deepnet in this PR?

ESCRI11 (approved):

PR #1513, opened 2025-07-24
feat: select raw/normalized pgx for MO

ESCRI11:
Added option to select raw or normalized data to be extracted from the selected pgx files (that will be fed into the multi-omics computation)

also reshufled the code so it makes sense, old approach was not OK too convoluted

image


ESCRI11:
> Suggestion: "Select data processing level": delete "data" from "Raw data" and "Normalized data" >> "Raw"; "Normalized"

I like the idea, changed it following your suggestion

PR #1521, opened 2025-07-28
Enable Olink NPX

zitoa:
Enhancement: enable upload and analysis of Olink NPX proteomics data. Attempted to achieve this by minimum code changes. Needs https://github.com/bigomics/playbase/pull/293

If proteomics datatype is selected, a dropdown menu will ask selection of MS vs Olink.
image

This very first PR allows upload of Olink NPX data format that strictly conform with the Olink standard format (validated by OlinkAnalyze R package). We could allow more flexibility afterwards.

No need to upload samples.csv. This PR includes automatic detection of Olink NPX format and generation of samples metadata from there. We could allow more flexibility afterwards.

If Olink NPX is detected, normalization is set to FALSE (NPX are log2 and normalized). Will be up to user to re-change that.

### DIFF
--- a/components/board.upload/R/upload_module_normalization.R
+++ b/components/board.upload/R/upload_module_normalization.R
@@ -14,34 +14,34 @@ upload_module_normalization_ui <- function(id, height = "100%") {
   uiOutput(ns("normalization"), fill = TRUE)
 }
 
+
 upload_module_normalization_server <- function(
     id,
     r_counts,
     r_samples,
     r_contrasts,
     upload_datatype,
+    is.olink,
     is.count = FALSE,
     height = 720) {
   shiny::moduleServer(
     id,
+
     function(input, output, session) {
       ns <- session$ns
 
       observeEvent(input$normalization_method, {
         shiny::req(input$normalization_method == "reference")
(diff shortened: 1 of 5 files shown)


components/board.upload/R/upload_module_normalization.R
```diff
@@ -250,7 +250,7 @@ upload_module_normalization_server <- function(
           out <- playbase::detectOutlierSamples(X, plot = FALSE)
 
           scaledX <- playbase::double_center_scale_fast(X)
-          corX <- HiClimR::fastCor(t(scaledX), optBLAS = TRUE)
```
ESCRI11, line 253: Why change this?

ESCRI11 (changes requested):
Resolve merge conflicts, then looks good

components/board.upload/R/upload_module_normalization.R
zitoa: Can't remember. I will check. Great it looks good

zitoa: Good catch. I reinstated HiClimR::fastCor(t(scaledX), optBLAS = TRUE)

PR #1522, opened 2025-07-29
fix: handle multiple genes with same symbol

ESCRI11:
Issue from customer dataset (salmon):

on 

```
gx <- X[which(xgene == pheno), kk, drop = FALSE]
```

we assume that this will yield a 1 row table, which in alternative species (non human/mouse) is not always the case, by `colMean` `gx` in the case it has more than one row we solve the issue

### Before
image

### After
image


zitoa:
Duplicated features are either averaged or made unique at upload. Do duplicates involve pgx$genes$symbol while pgx$genes$features are unique? If so your solution is ok

ESCRI11:
Exactly what you mention

zitoa (approved):

PR #1529, opened 2025-07-30
Rework heatmap and heatmap editor

ivokwee:
Rework of heatmap and editor:
- Labels were not seen if user selects (or inputs) more than 50 genes (this was noted by AZ@irb). Instead of auto-hiding, replaced with explicit show_rownames option.
- input$num_rownames and input$hm_ntop are confusing. Removing num_rownames.
- Dendrogram checkbox do nothing (in Editor);
- show_legend duplicated in hamburger and editor. Remove from Editor. 

NOTE1:  Editor does not work for dynamic heatmap (plotly)? 
NOTE2:  If permanent, commented out code in ui-EditorContent.R can be removed
NOTE3:  we need some rules of what to put in editor and what in the hamburger menu. 
NOTE4:  previously there was a choice option dynamic/static in the zoomed plot. It does not appear in most cases on my browser. Please check.

### DIFF
--- a/components/board.clustering/R/clustering_plot_splitmap.R
+++ b/components/board.clustering/R/clustering_plot_splitmap.R
@@ -35,6 +35,7 @@ clustering_plot_splitmap_ui <- function(
       ),
       "Show gene-sample heatmap or sample-sample correlation plot."
     ),
+    shiny::hr(),
     withTooltip(
       shiny::radioButtons(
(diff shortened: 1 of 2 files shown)


ivokwee:
> @ivokwee please commit the last changes

ok. I forgot to push. After merge please update the deploy (at least for IRB1).

PR #1535, opened 2025-08-06
Control MultiOmics

ESCRI11:
Feature: ability to control if MultiOmics is enabled for a user:

- Set MultiOmics default as FALSE
- Read MultiOmics option from DB and OPTIONS file
- If MultiOmics is disabled
  - Disable MultiOmics data upload
  - Grey out any MultiOmics datasets on your library


zitoa:
@ESCRI11 am not very updated with ongoing discussions about how to control MO. Better IK to review this @ivokwee

ESCRI11:
true

ivokwee:
OK. Nice. seems to work. 

image

PR #1538, opened 2025-08-12
Devel sc thld

ESCRI11:
If uploaded SC data has more cells than allowed, it goes through and we set the limit of supercell to max SC cells, that way data will go though.


zitoa:
We do SuperCell in pgx.createSingleCellPGX. limit to 10K. above that supercell is enforced...

zitoa (approved):

PR #1575, opened 2025-10-05
Make timeseries optional

zitoa:
Make timeseries optional

### DIFF
--- a/components/board.upload/R/upload_module_computepgx.R
+++ b/components/board.upload/R/upload_module_computepgx.R
@@ -39,6 +39,7 @@ upload_module_computepgx_server <- function(
     process_counter,
     reset_upload_text_input,
     probetype) {
+
   shiny::moduleServer(
     id,
     function(input, output, session) {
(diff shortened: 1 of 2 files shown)


zitoa:
Wait before review @ESCRI11 .  
it needs some other fix.

zitoa:
Ok should be ready for review now. Rationale: timeseries is complex. Only trigger if "time" is colnames(samples) and if user select "timeseries" and if any valid contrasts (not degenerate) present.

components/board.upload/R/upload_module_computepgx.R
```diff
@@ -428,77 +431,58 @@ upload_module_computepgx_server <- function(
...
+          for (i in 1:ncol(Contrasts)) {
+            ctx <- colnames(Contrasts)[i]
+            if (strsplit(tolower(ctx), ":")[[1]][1] %in% time.var) next
+            tt <- table(data.frame(ctx = Contrasts[, ctx], time = Y[, sel.time[1]]))
+            zeros.obs <- apply(tt, 1, function(x) sum(x == 0))
+            if (length(unique(timeseries)) == 1 && unique(timeseries)[1] == "") {
+              if (!any(zeros.obs)) ia.ctx <- c(ia.ctx, ctx)
+            } else {
+              if (!any(zeros.obs >= (ncol(tt) - 1))) {
+                ia.spline.ctx <- c(ia.spline.ctx, ctx)
```
ESCRI11, line 466: Please avoid things like that, for what I understand you just changed the indentation of this whole code, but if there is any change inside it I have no clue. This makes it very hard to review PRs

ESCRI11 (changes requested):
On example data, which contains "time" column on samples I do not get the time series selector

PR #1612, opened 2025-11-04
add_check_scrnaseq_datatype

zitoa:
Draft basic 1st check to protect wrong datatype selection between bulk vs. scRNA-seq.


zitoa:
Improved it, thanks.

PR #1620, opened 2025-11-11
feat: retain common samples on pgx multiomics upload

ESCRI11:
As discussed, retain the common samples when uploading multi omics data from pgx files.


ESCRI11:
@zitoa that should be handled by the already placed checks on samples step -- let me double check

PR #1624, opened 2025-11-13
Enable multispecies analysis

zitoa:
Enable multispecies analysis. Requires https://github.com/bigomics/playbase/pull/361

We have two requirements for any user interested in multi-species:
1.  Add an extra column 'species' or 'organism' in the counts.csv reporting official organism identifier for each feature.
2.  Select more than 1 "organism" at upload. 

For multi-species analysis, we append the species to the feature identifier. eg
GENE1_Human
GENE2_Rat


phisanti:
Shall we merge this? Is it waiting for further feedback?

ivokwee:
Does it handle correctly is an underscore is already present in the feature name? lipid names, metabolltes, but sometime already gene and protein id may use already undscores?

ivokwee:
this now needs to point to devel  (edgy is no more...)

Issue #1615, opened 2025-11-07
req: add custom filter in foldchange heatmap


ivokwee:
[Request from GQ@m, 7.11.2025]

I want to be able to selectively choose proteins to show on the Fold Change Heatmap.” I tried to use the filter button and look for a “Custom” but it won’t let me do so. Is there a way to show only a custom list of proteins for foldchanges  using this heatmap. Proteinset does not have a custom function either.

Possible solution:  Add gene filter as in 'Simlar Experiment/FC heatmap'.

<img width="1122" height="298" alt="Image" src="https://github.com/user-attachments/assets/d7b5b651-4434-409d-960e-edb1920875a8" />

PR #1626, opened 2025-11-18
feat: add custom gene filter to intersection heatmap

ESCRI11:
This closes https://github.com/bigomics/omicsplayground/issues/1615

- Add <custom> filter option to gene-level filters
- Add debounce to the input text
- Removed custom filter code from geneset level


ESCRI11:
Added

- Synchronize `qv` matrix reordering with `df` matrix to ensure correct significance coloring
- Add opacity control (10% for non-selected, 100% for selected genes)
- Remove unused variables and redundant sorting

PR #1637, opened 2025-12-02
cleanup: remove chirpbutton

ESCRI11:
Remove the `Discuss!` button. It goes to our reddit, which contains 2 posts from 4 years ago, we don't need a button that stays on the UI the whole time for that. https://www.reddit.com/r/OmicsPlayground/

image


zitoa (approved):
This was a long-standing problem....

PR #1673, opened 2026-01-08
feat: properly position shiny validate/need

ESCRI11:
From hubspot tickets

## Before
image

## After
image


zitoa:
Can we entirely remove the parallel plot? It is useless in my opinion (have been saying this for months).
What you think Xavi?

ivokwee:
> Can we entirely remove the parallel plot? It is useless in my opinion (have been saying this for months). What you think Xavi?

if you want in a new PR. But first ask other users, Axel for example.

PR #1678, opened 2026-01-13
Improvements multiwgcna

ivokwee:
Note: Needs  PR bigomics/playbase#420 on playbase. 

1. Added AI report and diagram for WGCNA and multiomics WGCNA.
2. Various fixes.


zitoa:
Multi-omics WGCNA should not appear in example-data (as well as any other single-omics data).
image

zitoa:
@ESCRI11 Exactly same crash i get...

ivokwee:
ok. mox-brca still has problems. Please try mox-geiger.pgx (proteomics + metabolomics). But the diagram still has often problems in the generated code.

ESCRI11:
Maybe we can catch failed diagram and re-send to model until something useful comes up?

ivokwee:
I added few manual corrections. It should be much less now. Also Llama is pretty poor for the diagram. Yeah, we could perhaps do a 2nd round like "check and correct this code" (could be cool). Or maybe igraph is easier for them?

ivokwee:
> also, if I compute multiWGCNA AI report and then go to standard WGCNA and try to get AI report I get errors 

OK. same error. Something in getTopModules. I will look at it.

zitoa:
Am testing again the PR. Where the corresponding playbase PR? @ivokwee

zitoa:
Mox-brca errors: 
image

zitoa:
Diagram seems to be generated by LLM directly....that is what happens

zitoa:
playbase PR used for testing https://github.com/bigomics/playbase/pull/418; dataset is mox-brca: crash in multiWGCNA:
image

zitoa:
just committed a push to fix a typo most likely causing an edge-case crash in the future.

zitoa:
mox-brca: crash 
image

zitoa:
I believe there are changes not committed or something went off. There are multiple missing parameters in functions resulting into "objects" or "parameters" not found from corresponding playbase PR. I pushed some fixes...more needed.

zitoa:
Loading fresh example data. Error:
image

ESCRI11:
on a separate note, ai diagram still fails sometimes, we need a mechanism to gracefully fail instead of displaying a random error message to the user (which everytime is different)

PR #1715, opened 2026-02-10
improvements on  multiwgcna (attempt 3)

ivokwee:
Various improvements on multiWGCNA

Note:  Needs Playbase PR  https://github.com/bigomics/playbase/pull/420


ivokwee:
> Screenshot 2026-02-12 at 09 46 30
> Cannot find the AI features, even when I turn on the `Enable Beta Features`

weird. please check the code.

ivokwee:
> at this moment ENABLE_CHIRP controls wheter the Copilot buttons appears or not. We can use this PR if you want to change it to control if the Enable AI selector appears.

Let me think. Perhaps the wgcna and reporting will be available for beta sooner than the copilot. Also we somehow need to detect if LLM  and/or image genAI (for infographic) are available by checking if any API_KEY is available. Boards should not crash if API_KEYS are not available.

zitoa:
@phisanti are you working on resolving these bugs (to avoid 2 devs working on same thing). ? Make sure you use the correct playbase branch. Let me know ASAP. If you are not resolving these bugs I will do it. Thanks.

phisanti:
> image Diagram is still not controlled properly it seems

This is likely due to the agent formatting the `DOT` diagrams. This is solved in `feat_summaries` + `omicsai` using `edgelist` format instead of `DOT`. The `edgelist` is more JSON-like and thus less error-prone.

zitoa (commented):
Why 2 samples per group or 2 samples in total are not enough? In this case just reports single MM value.

zitoa:
something is wrong in the plot.
image

ESCRI11:
image
gemini 3.1 flash decided it was a good idea to put nature logo, we have to add something to the promp to prevent this sort of stuff happening

PR #1745, opened 2026-03-30
feat: admin panel with user/data management

ESCRI11:
Adds admin panel with user and data management:
 - app stats overview
 - editable credentials table: password and admin condition
 - data: moving/deleting datasets across user folders (and shared bucket as well)

Admin stuff buried in two layers
1. OPTIONS file to load server for admin board
2. Per user credential option

To test:

on options file add: `ENABLE_ADMIN = TRUE`
on credentials file add column named "ADMIN" and set the user you want to test to TRUE


ivokwee:
Suggestions (not critical):
1. Selecting user, then deleteing 1 file, the user filter resets. Would be good to show still files of user.
2. Would be good to be able to delete all or selected files of user
3. Maybe add a confirmation dialog after pressing move. or have the destination dropdown in the 'move' popup.

ESCRI11:
@ivokwee is it fine if I merge and open issue with your suggestions for future improvement? -- I like them and feel they have to be included, just to move faster for now

ivokwee:
OK. I see multiple select actually works. 
- also would be nice to have  number of pgx per user in the user management table
-

ivokwee (approved):

Issue #1758, opened 2026-04-24
"Error: incorrect number of dimensions" for multiomics heatmap


ivokwee:
Plot gives "Error: incorrect number of dimensions" error with mox-daf2.pgx dataset. Error most probably in playbase::deep.plotBiomarkerHeatmap(). 

<img width="2613" height="1553" alt="Image" src="https://github.com/user-attachments/assets/2cdd6f28-9dc2-4e31-88ed-7154b08eb224" />

```
Warning: Error in [: incorrect number of dimensions
  201: h
  200: .handleSimpleError
  199: data.frame
  198: cbind
  197: cbind
  196: playbase::deep.plotBiomarkerHeatmap
  195: func2
  194: renderPlot
```

ivokwee:
fixed in PR #1760

PR #1760, opened 2026-04-24
add DROP=FALSE to fix error

ivokwee:
Add DROP=FALSE to fix issue #1758


ESCRI11 (approved):
hahaha, what a classic... :D

PR #1761, opened 2026-04-25
Add option to group by feature classes the boxplots

zitoa:
Add option to group by feature classes the boxplots. Only valid if phenotype is selected and "class" present in feature annotation. Needs https://github.com/bigomics/playbase/pull/464

Ask me for example data.


ESCRI11 (approved):
added an improvement for the plot editor of this new feature

PR #1766, opened 2026-05-06
feat: plot editor order umap subplots

ESCRI11:
Client request for publication

Enable order of subplots on umaps clustering>features

### DIFF
--- a/components/board.featuremap/R/featuremap_plot_gene_sig.R
+++ b/components/board.featuremap/R/featuremap_plot_gene_sig.R
@@ -46,9 +46,9 @@ featuremap_plot_gene_sig_server <- function(id,
     output$rank_list <- shiny::renderUI({
       dt <- plot_data()
       shiny::req(dt)
-      F <- dt[[1]]
-      shiny::req(F, ncol(F) > 0)
-      rank_list_ui(sort(colnames(F)), session$ns)
+      facet_df <- dt[[1]]
+      shiny::req(facet_df, ncol(facet_df) > 0)
+      rank_list_ui(sort(colnames(facet_df)), session$ns)
     })
 
     plot_data <- shiny::reactive({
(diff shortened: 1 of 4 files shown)


components/board.featuremap/R/featuremap_plot_gset_sig.R
```diff
@@ -45,6 +46,14 @@ featuremap_plot_gset_sig_server <- function(id,
                                             plotFeaturesPanel,
                                             watermark = FALSE) {
   moduleServer(id, function(input, output, session) {
+    output$rank_list <- shiny::renderUI({
+      dt <- plot_data()
+      shiny::req(dt)
+      F <- dt[[1]]
```
phisanti: This is not a good assignment. In R "F == FALSE". I will create confusion downstream and for reading.

phisanti: Tho, I know we do that everywhere 😆

components/board.featuremap/R/featuremap_plot_gene_sig.R
```diff
@@ -42,6 +43,14 @@ featuremap_plot_gene_sig_server <- function(id,
                                             plotFeaturesPanel,
                                             watermark = FALSE) {
   moduleServer(id, function(input, output, session) {
+    output$rank_list <- shiny::renderUI({
+      dt <- plot_data()
+      shiny::req(dt)
+      F <- dt[[1]]
```
phisanti: Once more, I discourage using assignments to `F <- foo()`.

phisanti: See the statement below `req(F, ncol(F) > 0)`, reads as `req(FALSE, ncol(FALSE) > 0)` while it behaves as `req(dt[[1]], ncol(dt[[1]]) >0)`.

phisanti (approved):

PR #1771, opened 2026-05-13
fix: guard against missing cols

ESCRI11:
with example-data

symbol col can miss if showall is not selected

### Before
image

### After
image


phisanti (approved):
It works/renders in my computer

PR #1772, opened 2026-05-13
feat: explain grey out datasets

ESCRI11:
per marketing team: add something to let user understand why a dataset is greyed out

https://github.com/user-attachments/assets/cbb274fa-997c-4120-865d-fec6ed672e9b


phisanti (approved):
We should count how many "on-the-fly" JS injections we do. It can become handy to have the snippets written + a read_js() utility.

PR #1775, opened 2026-05-18
fix: match hyperbolic volcano with expr genes table

ESCRI11:
from client feedback

when using the "hyperbolic" volcano plot, the table does not match

note the difference in table dimensions:

### before
image

### after
image

### DIFF
--- a/components/board.expression/R/expression_server.R
+++ b/components/board.expression/R/expression_server.R
@@ -293,7 +293,19 @@ ExpressionBoard <- function(id, pgx, labeltype = shiny::reactive("feature")) {
       if (!is.null(input$gx_showall) && !input$gx_showall) {
         n <- length(tests)
         ## sel <- which(res$stars == playbase::star.symbols(n))
-        if (input$show_pv) {
+        ct <- if (exists("volcano_cutoff")) volcano_cutoff$cutoff_type() else NULL
+        if (!is.null(ct) && ct == "hyperbolic") {
+          ## Match playbase::ggVolcano hyperbola: y uses pval_cap-floored
+          ## meta.p / meta.q on the same scale as the plot.
+          cap <- pval_cap()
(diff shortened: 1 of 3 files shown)


components/board.expression/R/expression_server.R
```diff
@@ -293,7 +293,19 @@ ExpressionBoard <- function(id, pgx, labeltype = shiny::reactive("feature")) {
       if (!is.null(input$gx_showall) && !input$gx_showall) {
         n <- length(tests)
         ## sel <- which(res$stars == playbase::star.symbols(n))
-        if (input$show_pv) {
+        ct <- if (exists("volcano_cutoff")) volcano_cutoff$cutoff_type() else NULL
```
phisanti, line 296: I would extract all reactives together (easy for clankers to forget in future refactors).

Something like:
```r
 if (!is.null(input$gx_showall) && !input$gx_showall) {
    n <- length(tests)

    ## --- Extract reactives ---
    ct  <- volcano_cutoff$cutoff_type()
    k   <- volcano_cutoff$hyperbola_k()
    cap <- pval_cap()

    ## --- Build significance mask ---
    if (!is.null(ct) && ct == "hyperbolic") {
      y <- if (input$show_pv) {
        -log10(pmax(res$meta.p, cap) + cap)
      } else {
        -log10(pmax(res$meta.q, cap) + cap)
      }
      sel <- which(hyperbolic_significance(res$logFC, y, psig = fdr, lfc = lfc, k = k))
    } else if (input$show_pv) {
      sel <- which(res$meta.p <= fdr & abs(res$logFC) >= lfc)
    } else {
      sel <- which(res$meta.q <= fdr & abs(res$logFC) >= lfc)
    }

    res <- res[sel, , drop = FALSE]
  }
```

phisanti (approved):
Again, approve with life quality comments.

PR #1780, opened 2026-05-29
Edgy fix copilot

phisanti:
## Summary

  - Reports auto-generated on every pgx load, no user consent, 2-3 min freeze.
  - Two observers fired on `pgx$X` change: one in `app_copilot.ivo`, one I almost added in loader.
  Conflict → "No" button ignored.
  - Single observer now lives in `LoadingBoard`. Prompts yes/no, runs `playbase::pgx.update_reports`,
  saves via `playbase::pgx.save` (user-dir only).

  ## Why here, not in copilot

  1. Load issue, not copilot issue. Trigger = "new pgx in memory" = loader's domain.
  2. Auto-generation centralized. One source of truth:
  `board.loading/R/loading_server.R::maybe_offer_ai_reports`.
  3. One observer decides. No race between copilot + loader.

  ## Changes

  - `components/board.loading/R/loading_server.R` — new `maybe_offer_ai_reports(pgxfile, is_user_dir)`
  helper. Called from `loadAndActivatePGX` after pgx pushed to `reactiveValues`. Skips if no LLM
  configured or any report slot already populated. On "Yes": update + patch slots back + `savePGX`
  (gated on `is_user_dir`).
  - `components/app_copilot.ivo/R/copilot_server.R` — drop `pgx.update_reports` block from dataset
  observer. Section checkbox refresh kept.


ivokwee (approved):

ivokwee:
great addition. Perhaps should also do it for infographics?

PR #1783, opened 2026-06-01
Edgy cleanup

phisanti:
## Summary

Cleanup branch off `edgy`. Two refactor commits, zero behavior change at default config.

- **11 → 1 `%||%` definitions.** 9 duplicate shims across `components/app_copilot/R/*` removed. Canonical def stays at `components/app/R/utils/utils.R:18`. `playbase::%||%` also exported. Local `CLAUDE.md` notes added (not tracked — local convention).
- **max_history single source.** `copilot_save_controller` was inline `getOption("copilot.max_history", 20L)`; helper `copilot_max_history()` defaults to 100. Save controller now calls helper. No more silent 20-vs-100 split.


ESCRI11 (approved):
added a default to global.R files for `LLM_MAXTURNS` to avoid crashes in case we forget it on the OPTIONS file (should not happen but happened in the past with other arguments), little extra safety for free

PR #1784, opened 2026-06-02
feat: add answer-style knob to copilot chat (concise / technical / custom)

phisanti:
## Summary

- Copilot chat now has an **Answer style** radio alongside the existing model-tier radio: **Concise**, **Technical**, and **Custom** (free-text instructions).
- The style applies to the next user message via `ellmer`'s `set_system_prompt()` and persists with the saved session, so reloading a chat keeps the chosen style.
- Model tier shrinks from 3 options to 2 (**Balanced** / **Deep**); the old `copilot-fast` value still resolves to Balanced so saved sessions don't break.

## Why

Users reported answers were too verbose. An internal A/B comparison showed the strict "concise" preset cut output ~50% on the same questions, so we expose it as a first-class control rather than baking one tone in.

## NOTES
- Requires lates `omicsagentovi`
- in case we want to edit the output style, we only need to edit `omicsagentovi/inst/prompts`


ESCRI11 (approved):
lgtm

PR #1785, opened 2026-06-03
feat(upload): toggle full vs top-1000 features in QC/BC batch-correct

ESCRI11:
from client requests

add toggle to use all features for the BC PCA plots

image


components/board.upload/R/upload_module_normalization.R
phisanti, line 1: use `nrow(x)` instead of `Inf` can be safer.

phisanti (approved):
Approved with minor comments

ivokwee:
I think this upload QC/BC board earns the chicken-choco-pizza award...

chicken-choco-pizza

ESCRI11:
And they also requested more plots for the pca....

PR #1792, opened 2026-06-12
chore: remove buy us a coffee button

ESCRI11:
image

no more coffees


phisanti (approved):
Despite the fact that we are going to lose a significant source of caffeine, I think we are now at the point to do this step.

PR #1794, opened 2026-06-15
Feat/ai reports

phisanti:
## Summary

This PR brings the new AI report flow into OmicsPlayground.

AI reports are now stored under the PGX AI schema:

`pgx$ai[[report_module]]`

The supported report modules are:

- Differential expression
- Enrichment
- MOFA
- WGCNA
- Multiomics WGCNA
- Combined report

Reports can now be created in three places:

1. At compute time, through the new `playbase::pgx.update_reports(..., ai = ...)` path.
2. At load time, where the user is asked whether reports should be generated in the background while they start exploring the dataset.
3. In AI Studio, where users can regenerate selected reports and infographics.

AI Studio now reads and writes reports through `pgx$ai`. Infographics are also generated from the new omicsai image flow and stored back into the matching `pgx$ai` report slot.

Obi-two now supports different writing styles and can answer with or without tools. Precomputed AI reports can be passed as context, so the assistant can use report-level interpretation without regenerating reports.

Board-level AI Report tabs are now hidden. Some targeted AI summaries remain, including the WGCNA per-module summary, which is generated on demand from the current WGCNA result.

## What this PR does not remove

Legacy report-generation functions are still present. They are unwired from the active app flow, but kept in the codebase for now so we can compare behavior if needed. Full cleanup should happen in a follow-up PR after this migration is accepted.

## Dependency notes

This PR expects the companion package changes for:

- `playbase >= 4.1.01` 
- `omicsai >= 0.3.1`
- `omicsagentovi >= 0.5.75`
- `omicspgxmcp >= 0.4.49`


phisanti:
Linked PR in playbase: https://github.com/bigomics/playbase/pull/482

ESCRI11 (approved):

PR #1795, opened 2026-06-16
fix(correlation): decouple table search from plots, add row-select dr…

ESCRI11:
bug fix + client request

correlation table's search/sort modified scatter grid and the barplot (through `rownames_current()`), quite strange behaviour. also when filtering just for one feature the table, the scatter plot displayed was just the selected gene (on settings right bar) against itself (correlation 1)

added the option to click on the table to select which feature to show only on the scatter + highligted it on barplot

after_selected


phisanti:
lgtm

phisanti (approved):

PR #1804, opened 2026-06-23
Across pgx tiledb edgy

ESCRI11:
first iteration of across datasets feature

requires: https://github.com/bigomics/playbase/pull/487

image

image

image


phisanti:
Looking at how the new `apps` in the edgy structure/multi-app approach. I think we should move from `components/board.across` to `components/app_across`. That is at least the pattern for:
```
components/app
components/app_copilot
components/app_copilot.ivo
components/app_prism
components/app_studio
components/app_tools`
```

What do you think? This should be very mechanical, a pure `mv -R ...`

ESCRI11:
@phisanti good point, this feature was originally developed un devel branch so the move to edgy was taking into account old structure, will move asap

phisanti:
Created the DB for my data folder. 38 datasets, ~2GB -> 5min in my computer. This might take longer for users. I would move the first-time build of the DB from log-in to session to opening the new across-dataset tab (opt-in) and show message (shinyvalidate): `Updating the dataset db this takes a few minutes. Proceed? Yes/No`. Another option is that we run this process for all users overnight.

ESCRI11:
thats quite a good point, i agree (with first part, overnight compute could be overkill)

phisanti:
Minor, I would rename `Type Genes` and `Selected Genes` for `Type Features` and `Selected Features`. Given the fact that this acts across datasets, we cannot assume nor infer whether we are dealing with protein, genes, metabolites...

phisanti:
Minor/medium:

Gene list/Feature set does not "flush" when reloading a new subset of dataset. Reporduce:
1. Round 1:  Select 2 datasets, celegans-wormbase and klidel-multiomics which are _C. elegans_
2. Round 2:  Deselect those 2 and select human datasets example-data and example-data-mini, genelist adds the new human genes (they can be found) but the old items remain there.
Screenshot 2026-06-24 at 10 50 01
Another thing that might be happening is that the genelist provided corresponds to the whole feature set avaliable across all datasets. If that is the case, I would then filtrate the options shown to the datasets selected (if any). 

If implementation is too complex, this can be dropped.

phisanti:
> Another thing that might be happening is that the genelist provided corresponds to the whole feature set avaliable across all datasets. If that is the case, I would then filtrate the options shown to the datasets selected (if any).
> If implementation is too complex, this can be dropped.

Just noted that the pre-filtering is done within the Phenotype drop-down menu. Thus, this suggestion is just apply the same logic to the gene list box.

phisanti:
Screenshot 2026-06-24 at 11 00 41

User does not know what TileDB. I would reword as "View feature values for selected features across all samples from your datasets." or "Compare counts of selected features across every sample in your datasets, side by side.".

Note that I also dropped "Gene expression" in favour of "features".

phisanti:
Screenshot 2026-06-24 at 11 04 39


Here, I would:
1. include an extra column with datatype (metabolomics, proteomics...) and maybe organism.
2. Rename Gene column to Feature.
3. Update the "counts" colname if the user selects z-score of log2.

Another option could be to present our default but have a selector: choose columns.

Minor/non-blocker

phisanti:
Screenshot 2026-06-24 at 11 31 13


For this type of plot (faceted stacked as cloumn), can be useful to show the x-axis ONLY on the bottom row.

phisanti:
Screenshot 2026-06-24 at 11 42 22


After creating the tilledb for the very first time, it seems that the values are not refreshed for the board, and does it comes blank. This can be solved by logging out and logging in again. This suggests the bug comes from the fact that the code in component/app_across is run before the DB exists and does not refresh (my guess, not clanker, so probably wrong).

This can be reproduced by
```
rm -rf data/counts_tiledb/
rm -rf data/counts_tiledb_metadata.rds
```

Then log in again; the DB is generated, and click on the across App.

PR #1808, opened 2026-07-03
Chore/edgy cleanup

phisanti:
## Summary
Removes the old copilot and moves the last LLM/model calls off playbase onto omicsai, now that a single copilot (Obi) and AI Studio replace the old stack. Also fixes a report-context bug found while testing.

Depends on playbase `chore/edgy-cleanup` (https://github.com/bigomics/playbase/pull/491). Merge the playbase PR first.


ivokwee:
OK. Let's merge and cleanup in 2nd iteration:
- Cleanup/minimize prompts
- Remove organisms list if not needed
- UI changes in Obi

PR #1809, opened 2026-07-08
feat: send raw counts + preprocess settings to createPGX (bulk upload)

ESCRI11:
Depends on bigomics/playbase#493 — merge and rebuild playbase first.

The upload flow used to precompute the normalized matrix X in the normalization
module and pass it to `pgx.createPGX(X=...)`; a script/endpoint that only sent raw
counts got a plain log2 instead, so results diverged. The bulk path now sends RAW
counts + a serializable `preprocess` settings list, and createPGX rebuilds X via
`playbase::pgx.preprocess` — one implementation for app / script / endpoint.

- `upload_module_normalization.R`: expose one `preprocess` options reactive from the
  module's own inputs. Preview panels unchanged.
- `upload_server.R`: thread `preprocess` + the same RAW counts/annot the module
  consumes into the compute module.
- `upload_module_computepgx.R`: bulk → counts=raw, annot=raw, countsX=NULL,
  preprocess=opts; scRNA → unchanged (X is unused by createSingleCellPGX). countsX()
  still feeds DE method-selection and NA warnings.
- `pgxcreate_op.R`: pass `preprocess=params$preprocess`.

scRNA-seq path untouched. Verified with a Shiny `testServer` harness: the module's
real computed X equals `pgx.preprocess(raw, module$preprocess())` exactly across 9
setting combinations (norm methods, normalize-off, outlier removal, zero-as-NA,
imputation, missingness filter).

### DIFF
--- a/bin/pgxcreate_op.R
+++ b/bin/pgxcreate_op.R
@@ -25,6 +25,7 @@ pgx <- playbase::pgx.createPGX(
   organism = params$organism,
   counts = params$counts,
   X = params$countsX,
+  preprocess = params$preprocess,
   norm_method = params$norm_method,
   samples = params$samples,
   contrasts = params$contrasts,
--- a/components/board.upload/R/upload_module_computepgx.R
+++ b/components/board.upload/R/upload_module_computepgx.R
@@ -10,6 +10,9 @@ upload_module_computepgx_server <- function(
   id,
   countsRT,
   countsX,
+  preprocess = shiny::reactive(NULL),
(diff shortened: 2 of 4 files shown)


ivokwee:
Should also be exposed/copied to pgx$settings or somehow shown in user settings and reports.

ESCRI11:
good point, let me work on it

phisanti (approved):
Most of the issues addressed in other PR. This can merge sequentially.

PR #1812, opened 2026-07-09
feat: bring-your-own-key AI provider settings

phisanti:
## Summary
- add a Settings "AI Features" tab where admins curate allowed providers and each user picks a provider + supplies a session-only API key, threaded per-request into AI reports, image generation and the Copilot agent so enterprise users can run their own OpenAI/Anthropic/Google/GitHub/Mistral keys
- drive the per-provider model menus from a JSON selection policy (etc/ai_model_policy.json) layered over omicsai's live catalog, so model facts stay owned by omicsai while OPG owns only include/exclude/prefer curation
- gate every AI surface behind a licence flag (ENABLE_AI) and a runtime, admin-lockable usage switch: hide AI Studio/Obi and block the WGCNA module summary when off, and recover Copilot after a provider stream error with a friendly quota/auth message instead of the raw provider error

## Dependencies

Requires **omicsai ≥ 0.3.3** (provider-catalog + model-discovery API) and
**omicsagentovi ≥ 0.5.83** (catalog-resolved copilot tiers, BYOK credential
threading). Both must be released/installed alongside this branch.

https://github.com/bigomics/omicsagentovi/pull/1
https://github.com/bigomics/omicsai/pull/2


phisanti:
Basic view
Screenshot 2026-07-10 at 08 45 15

How setup work
Screenshot 2026-07-10 at 08 45 27

Supported providers (most common + custom endpoint if openai compatible)
Screenshot 2026-07-10 at 08 45 41

ESCRI11 (changes requested):
only relevant point found with clanker assisted review

Gating is inconsistent across AI surfaces. WGCNA summary and AiTextCardServer got explicit ENABLE_AI guards, but the auto-report paths in loading_server.R (maybe_offer_ai_reports) and upload_module_computepgx.R still gate only on llm_model being non-empty — not on opt$ENABLE_AI. The settings observer writes llm_model whenever input$enable_ai is TRUE, without re-checking the deployment licence (opt$ENABLE_AI). On an unlicensed deployment the switch is disabled but its value stays TRUE, so llm_model can still be written and auto-reports could fire. Worth confirming these two paths also short-circuit when !opt$ENABLE_AI

ESCRI11 (approved):
lgtm

PR #1813, opened 2026-07-09
Bring back dataset-sharing notifications on edgy

ESCRI11:
The dataset-sharing feature (notification + navigation + navbar badge when someone shares a dataset with you) lives on **devel**, but the **edgy** UI reshuffle broke it. This brings it back.

**What edgy broke:**
- The "New dataset received!" popup's *Go to shared datasets* button called `bigdash.selectTab(session, "sharing-tab")`, but that bigdash board was folded into the Library board's inner tabset — so the button did nothing.
- The pending-count badge died twice over: `ui.R` dropped the `<script src="static/shared-badges.js">` tag, and the badge JS still targeted the old bigdash navbar (`a[data-target="sharing-tab"]` / "Datasets" dropdown) that no longer exists.

**Fix:**
- Navigation now delegates to `LoadingBoard` → jumps to Library board + `sharing_tab` inner tab.
- Badge JS retargeted to the **Library** sidebar pill and the **Shared datasets** tab title; re-added the dropped script tag. CSS `.shared-pending-badge` was already present.

### DIFF
--- a/components/app/R/ui.R
+++ b/components/app/R/ui.R
@@ -52,6 +52,7 @@ app_ui <- function(x) {
       shiny::tags$head(shiny::tags$script(src = "static/copy-info-helper.js")),
       shiny::tags$script(src = "custom/close-message.js"),
       shiny::tags$head(shiny::tags$script(src = "static/add-tick-helper.js")),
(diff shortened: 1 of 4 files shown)


ivokwee (approved):

ivokwee:
this JS stuff... so ugly

PR #1815, opened 2026-07-13
fix: include feature name in data view expression CSV/Excel export

ESCRI11:
client request

Data view expression plot downloads (violin/box/barplot) were missing the feature name in the exported CSV/Excel — you got value/sample/group columns with no indication of which gene/protein they belonged to


phisanti (approved):
lgtm

PR #1806, opened 2026-06-24
feat: add AI usage telemetry

phisanti:
## Summary

This PR will allow us to monitor token usage per user with two db:
- data/user@email.sql
- etc/ai_telemetry.sql

The main logic lives in `components/modules/AiTelemetry.R`.

- `ai_telemetry_collect_chat():` runs on the copilot save tick, gather token usage per turn.
- `ai_telemetry_record()` writes token usage in db.

## Verify

Add:
```
USER$username <- "test"
USER$email <- "test@bigomics.com"
```

1. Start new session with Obi-Two. Send message. New db should appear with token count.
2. Repeat operation, new rows added.
3. Reopen old session, new rows added*.

*Note, token usage comes per turn to avoid issues with reopening sessions. Aggregation should be done in our side.

## Dependency

Merge omicsagentovi `feat/session-store-db-name` first.

PR #1816, opened 2026-07-13
feat: record deduplicated token telemetry for AI reports

phisanti:
## What

Copilot chat token usage is tracked (#1806), but AI **reports** spent tokens
invisibly. This records one deduplicated, priced token event per report
generation, across all generation sites.

**NOTE**:
Depends on  **playbase `feat: thread AI-report token usage onto the pgx report slot`** (PR https://github.com/bigomics/playbase/pull/497)

## Changes

- **`ai_telemetry_record_reports()`** (`AiTelemetry.R`) — reconcile helper that scans `pgx$ai` and emits one event per report keyed `report:<dataset>:<module>:<created_at>`. `INSERT OR IGNORE` collapses replays across sessions/reloads; NULL-usage slots are skipped; fire-and-forget so a telemetry failure never breaks report display.
- **Three generation checkpoints wired** — AI Studio, upload finish, and the loading generation path — each records from the persisted slot.
- **Manual edits marked** (`edited`/`edited_at` in `ai_report_update_text`) so a save never mints a new token event; `created_at`/`usage` are untouched.

## Testing

- `test-ai-telemetry.R` (35) + `test-ai-reports.R` (92) green, including a regression guard asserting the token columns populate (not just row count).
- Reviewed: field-name contract, `event_id` idempotency, `created_at` round-trip, slot-swap preservation, fire-and-forget isolation — 0 issues.

### DIFF
--- a/components/app_studio/R/aireport_server.R
+++ b/components/app_studio/R/aireport_server.R
@@ -563,17 +563,12 @@ AiReportServer <- function(id, pgx, save_pgx = NULL) {
 
           report_jobs$done <- report_jobs$done + 1L
 
-          # Telemetry: log one event per report generation on the main thread.
-          # TODO(telemetry): tokens unavailable until playbase returns usage from
-          # pgx.update_reports() (computed in the future worker, not returned here).
-          # TODO(telemetry): thread auth$email for attribution (no auth in StudioServer scope).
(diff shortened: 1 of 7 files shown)


ESCRI11 (approved):
lgtm

PR #1821, opened 2026-07-16
fix(dataview): stop the "type for more genes" hint from being selectable

ESCRI11:
from marketing team:

A client selected "(type for more genes...)" in the DataView gene picker and got broken plots — it was a genuine selectable option

Kept the hint (clients need it), just made it unclickable:
- greyed + pointer-events:none so it can't be clicked


phisanti (approved):
lgtm

PR #1823, opened 2026-07-20
Fix/chatbot UI

phisanti:
## Summary

Reshuffle of the Chatbot UI:

- Right column now stacks Dataset/Settings/History above the plot-rendering Evidence card (left column keeps just Docs/Reports context).
- Consumed reports stay full-colour instead of greying out — locked via `pointer-events`, with a lock icon + tooltip ("reports can only be passed once per session").

```
components/app_copilot/R/
├─ CopilotBoardUI.R         # column relayout
├─ CopilotReportsServer.R   # lock-not-grey checkbox + tooltip
└─ copilot_messages.R       # +report_locked message
```

### DIFF
--- a/components/app_copilot/R/CopilotBoardUI.R
+++ b/components/app_copilot/R/CopilotBoardUI.R
@@ -18,10 +18,27 @@ CopilotBoardUI <- function(id) {
     col_widths = c(3, 5, 4),
     style = "height: calc(100vh - 80px);",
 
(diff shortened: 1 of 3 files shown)


phisanti:
Screenshot 2026-07-20 at 14 58 56

ESCRI11 (commented):
lgtm

PR #1825, opened 2026-07-22
fix: gate AI report/infographic generation on dataset ownership

phisanti:
## Summary 

This PR adds cost monitoring to the AI reports and infographics. It also allows only admins can regenerate AI reports. This second part will need to be carefully checked that it aligns with standard OPG practices.


ESCRI11 (approved):
lgtm

PR #1830, opened 2026-07-28
feat: show normalized data in Sample QC, not raw counts

ESCRI11:
Sample QC read `pgx$counts` in all five plots, so it always showed the raw
upload — normalization and batch correction were invisible on the one tab
where you go to check them.

New **Sample QC data: normalized | raw** toggle in the settings panel,
defaulting to normalized (`pgx$X`, linearized so totals and proportions
still mean something).


phisanti (approved):
lgtm

PR #1831, opened 2026-07-28
fix: stop rejecting uploads over their filename

ESCRI11:
Stop rejecting uploads over their filename

Uploads no longer need "counts"/"samples"/"comparison" in the name. The check was added when the name picked the read path (IS_COUNT/IS_EXPRESSION); that code is long gone. Extension checks and the params.RData devmode branch stay.


phisanti (approved):
lgtm

PR #1834, opened 2026-08-04
feat(basic-menu): flatten the menu and lock down advanced settings

ESCRI11:
Extends the **Basic menu** switch (App settings) so it actually simplifies the boards, not just the sidebar. Until now it swapped `#menu-full` for `#menu-basic` and nothing else — every board still showed its full settings panel and all its tabs.

Base is `edgy`, not `master`: `opg_ui.R`, `opg_server.R` and `scss/components/_app.scss` do not exist on `master` (which is 328 commits behind), and the `menu_basic` switch itself only lives here.

## Changes

**Menu — three entries, flat**

Basic mode now shows DataView, Cluster Samples, Differential expression. Geneset Enrichment and Pathway analysis were dropped from the subset.

`createMenu()` only emits a flat sidebar item when the tree key matches the tab name (case-insensitive); the remaining entries were keyed by title, so `clustersamples` and `diffexpr` each rendered as a collapsible group wrapping one identical child ("Cluster Samples > Cluster Samples"). Keys now match the tab ids.

**Settings greyed out in basic mode**

`.advanced-option` marks the blocks to lock:

| Board | Greyed |
|---|---|
| DataView | Options accordion (Scale, Sample QC data) |
| Cluster Samples | Filter options + Advanced options accordions |
| Differential expression | FDR + logFC, Options accordion |

Geneset Enrichment (FDR + logFC, Options) and Pathway analysis (Options) keep their `.advanced-option` markers too. They are inert while those boards sit outside the basic menu — the markers are keyed on the `body.basic-mode` class, not on the menu, so they stay correct if either board comes back.

Greyed accordions are force-expanded — the settings are locked, but the user should still see which ones are in effect.

**DataView and Differential expression show only their Overview tab.**

## Notes for the reviewer

- **Why CSS and a `body.basic-mode` class rather than `shinyjs::disable()`** — boards are inserted lazily, on `input$nav`, i.e. after the toggle observer has already run. A `disable()` at toggle time would be lost for any board not yet in the DOM. A class rule applies to whatever gets inserted later, for free.
- **Why `!important` on the hidden tabs** — `tab_control()` calls `shiny::showTab()`, which writes an inline `display`. A plain class rule loses to it. The accordion rules need no `!important`: `body.basic-mode .advanced-option .accordion-collapse` (0-3-1) already beats bootstrap's `.collapse:not(.show)` (0-2-0).
- Switching to basic also resets `dataview-tabs` and `diffexpr-tabs1` to Overview, since the active tab may be one that is about to be hidden.
- `styles.min.css` is regenerated from `scss/components/_app.scss` via `Rscript dev/sass.R`; verified byte-identical to the committed source.

## Explicitly not included: skipping board loads

The first cut of this branch also skipped inserting board UI for tabs unreachable in basic mode, aiming at a ~2 min load on Geneset Enrichment. That was based on a wrong premise and has been dropped.

Shiny suspends outputs in hidden elements (`suspendWhenHidden` defaults TRUE), and bigdash's `toggleTab` fires the `hidden` event that drives it. The sibling boards in a module (Signature, WordCloud, …) sit in hidden `.big-tab` divs and **never render while their tab is inactive** — there is no rendering left to save. The remaining cost is the visible board's own outputs, which the user needs.

An independent review of that dropped code also found two regressions it would have shipped, both now moot: `inserted_tabs` was never cleared when `bigdash.removeTab` deletes a module's tabs (blank boards after a transcriptomics → multi-omics → transcriptomics round-trip), and boards skipped in basic mode lost their `update*Input` messages, so toggling back left their selectors empty.

If the Geneset Enrichment load time is worth chasing, it needs profiling of that board's own plots and tables — a separate change.

## Testing

Static only — the app was not launched in this environment. Verified: all three basic entries satisfy `createMenu`'s flatten condition; `class=` lands as an attribute on both `bslib::accordion` and `bslib::layout_column_wrap`; the compiled CSS matches its source. **Worth a manual pass on the toggle before merging.**


ESCRI11:
### Basic menu is now admin-configurable

**1. Admin panel > Basic Menu.** Two lists: which boards the flat menu keeps, and which boards get their advanced settings greyed. Saved per deploy to `etc/BASIC_MENU-<HOSTNAME>` / `etc/BASIC_LOCKED-<HOSTNAME>` — etc/ is a shared S3 mount, so the hostname keeps servers apart. Applies on page reload.



**2. Greying is automatic now.** Every accordion in a settings sidebar is marked at the single point where board tabs are inserted, so 6 hand-placed marks are gone and the ~9 boards nobody had marked are covered too. Below is Correlation analysis, which we never touched. An accordion stays live if it holds an action button (WGCNA's "Recompute!") or opts out with `keep-live`.



**3. `FORCE_BASIC` pins a user to basic mode** — per user (OPTIONS file or DB row) or whole deploy. The switch is on and disabled.



30 tests in `tests/testthat/test-basic-menu.R`. Also in this branch: a regenerated `00SourceAll.R` — the committed one omitted `AiTelemetry.R` and listed 10 dead files, so the app died on login. Separate commit.

<sub>Screenshots live on the `assets/pr-1834` branch (images only, no code).</sub>

ESCRI11:
### Admin panel now uses the Settings layout

`shiny::tabsetPanel` (tabs across the top) → `bslib::navset_pill_list` with the nav
list down the left, same shell as the Settings board, so the two read as one app.





Also since the last update: greying no longer resets board tabs before any board
exists, `BASIC_LOCKED` is documented in `etc/OPTIONS` as a deployment default, and
the branch was re-verified against merged bigdash master (#41) — 21 assertions
across 5 configurations and 8 browser sessions, plus 30 unit tests.

<sub>User Management and Data Management use the same layout; not shown here as they render account data.</sub>

PR #1838, opened 2026-08-12
PCA: select components PC1-PC5, biplot arrows, phenotype directions and variance plot

ESCRI11:
Client asked to see up to PC5 and choose which component is plotted against which, plus a biplot, phenotype directions and a variance plot.

`pgx$cluster$pos` only stores PC1-PC3, so the board now recomputes PC1-PC5 on demand. Everything is gated to the `pca` layout — t-SNE/UMAP axes are unordered, so "PC3 vs PC5" means nothing there.

### Pick the components — sidebar, under `Layout`
component selectors

Both panels follow the choice:
PC2 vs PC3

### Biplot controls — plot options
biplot controls

### Arrows: `features` = gene loadings, `phenotypes` = direction each phenotype increases
biplot
phenotype directions

### `variance explained` swaps the panel for a scree plot
variance explained

## Notes

- **Variance % are lower than before** — now divided by total variance instead of the 50 components irlba retains. The old numbers were inflated.
- **The image needs rebuilding**: `plotly.repel` added to `docker/Dockerfile.update`. Without it the repel toggle silently does nothing (degrades to fixed labels, does not break).
- No playbase change needed.

## Testing

43 unit tests on the new helper (prcomp agreement, sign stability, non-finite input, degenerate matrices, 2/3/6/8/13 samples). Full suite: no regressions. Driven end to end in a browser — the screenshots above are from that run.

An independent review found 6 defects in rendering and error handling, all fixed in `5327d95dc`. The browser captures found a 7th: `grey30` is a valid R colour but not valid CSS, so plotly silently drew the arrows pink (`885a0c8cc`).

## Known gaps

Image export renders labels unrepelled; X and Y can be set to the same component; CSV export ignores the variance view.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -17,6 +17,7 @@ Suggests:
     omicspgx,
     omicspgxmcp,
     omicsplots,
+    plotly.repel,
     testthat (>= 3.0.0)
 Config/testthat/edition: 3
 RoxygenNote: 7.3.2
@@ -35,6 +36,7 @@ Imports:
     forcats,
     future,
     ggforce,
+    ggrepel,
     ggVennDiagram,
     heatmaply,
     htmltools,
--- a/components/board.clustering/R/clustering_plot_clusterpca.R
+++ b/components/board.clustering/R/clustering_plot_clusterpca.R
@@ -261,7 +261,7 @@ clustering_plot_clustpca_server <- function(id,
           plt <- plt %>% plotly::add_text(
             x = pos[, 1], y = pos[, 2], text = rownames(df),
             textposition = "top center",
-            textfont = list(size = 12 * cex, color = "grey30"),
+            textfont = list(size = 12 * cex, color = "#4D4D4D"),
             showlegend = FALSE, hoverinfo = "skip", inherit = FALSE
           )
         }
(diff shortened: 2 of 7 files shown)


phisanti:
Screenshot 2026-08-19 at 11 47 11

Testing now. For the "variance explained" plot, I will add a dropdown menu called Plot type. Otherwise, it takes some time to figure out where it is. This code patter is what we already do in other cards where we have the same plot type. Another option is to place it on the siderail settings menu "Layout" == umap/pca/variance-pca?

phisanti:
Repel labels seems to do nothing?

phisanti:
Finally, I would use diferent colour for the text related to the arrows and the text related to the labels.

phisanti (approved):
Some minor comments related to aesthetics and organisations.

ESCRI11:
Thanks @phisanti — all three comments taken on board, fixed in `48938dfbf`. Screenshots below are before/after from two live instances (`885a0c8cc` vs the fix), same dataset and same PC1 × PC2, so only the thing under review differs in each pair.

### 1. "Plot type" dropdown

You were right that it takes a while to find — it was a checkbox sitting *seventh*, below the arrow settings. It's now a `Plot type:` dropdown leading the panel, following the same pattern as `wgcna_plot_enrichment.R` and `biomarker_plot_decisiontree.R`.

plot type dropdown

I went with the in-card dropdown rather than the sidebar `Layout == variance-pca` option: `Layout` also drives the phenotype panel next door, which has no variance view to show, so a fourth entry there would blank a panel.

The arrow controls now hide under the variance view — they have nothing to act on there:

variance view

### 2. Different colour for arrow text vs label text

They were `#333333` and `#4D4D4D` — three shades of one grey, so the biplot overlay and the sample annotation read the same. Arrows, heads, labels and repel leaders now take `omics_colors("terra_cotta")`, saturated against the desaturated `muted_light` point palette; sample labels move to `omics_colors("super_dark_grey")`.

arrow colour

### 3. Repel labels seems to do nothing

No code change here — it isn't a bug in the branch. Two things stack up:

- **The package is missing on your instance.** `plotly.repel` is gated behind `requireNamespace()`, so it degrades silently. `docker/Dockerfile.update` got the install line in `cfcc72098`, but **the image still needs rebuilding** — that's the caveat in the PR description.
- **The default is 3 arrows.** Three labels essentially never overlap, so a *working* repel moves nothing. It only looks alive at a higher arrow count.

Both screenshots above are at 20 arrows, where repel visibly separates the labels — so the mechanism is confirmed working once the package is present. Worth re-testing after the rebuild, at a higher arrow count than the default.

---

Also confirmed in the browser that the arrows render terra cotta rather than the pink an unparseable colour string produces (the `885a0c8cc` failure mode) — `plotly_build()` can't catch that, only a real render can. 43/43 helper tests still pass.

PR #1839, opened 2026-08-13
fix(ui): align static/dynamic card header with other cards

ESCRI11:
Cards with the `dynamic`/`static` toggle rendered a **44px** header; every other card is **33px**.

**Before**
before

**After**
after

## Cause

bigdash splices bslib's nav-pills bar into `.plotmodule-header` and strips its bslib classes with a literal gsub on `"card-header bslib-navs-card-title"` (`bigdash/R/plot-module.R:373`). Current bslib emits `card-header bslib-gap-spacing bslib-navs-card-title`, so the gsub misses, `card-header` survives, and a full 33px card header lands inside the plot header.

One rule in `scss/components/_cards.scss` neutralises it — class-order independent, so further bslib churn can't re-break it. Fixing the gsub upstream in bigdash would also work; worth doing separately.

## Testing

Driven in Chromium/Playwright on Differential expression and Clustering → Features. Header measured 44px → 34px, siblings 33px, same session with only the rule toggled. The residual 1px is the nav's line box; below the noise floor.

Independent review found no blockers.


ESCRI11:
Superseded — fixing this upstream in bigdash instead, so every consumer gets it (bigdash ships the same `html .card-header{height:33px}` in its own stylesheet).

PR #1840, opened 2026-08-13
fix(upload): duplicated features silently dropped; batch-correction method reset

ESCRI11:
Two bugs in the upload flow, both pre-existing and independent of the preprocess refactor.

## 1. Duplicated features were silently dropped

`cleanX` realigned counts to X with `intersect()`, which returns **unique** names — so indexing by it kept only the first row per duplicated feature. It also ran unconditionally, reducing data even when nothing needed realigning.

On data we already ship:

| dataset | features | duplicated | before |
|---|---|---|---|
| `salmon-ENSOKI` | 999 | 30 | **969** |
| `horse` | 31217 | 14462 | **~16755** |

This matters now because on edgy the compute path sends raw counts + `preprocess` and `pgx.createPGX` rebuilds X via `playbase::pgx.preprocess()`, which does **not** have this bug. Preview and compute disagreed — and `countsX` still drives default DE-method selection and the NA warnings.

Fix: realignment extracted into `realign_counts_to_X()`, mirroring `pgx.preprocess` — realign only when rownames differ, `%in%` subset when duplicated.

## 2. The chosen batch-correction method was reset to ComBat

`updateSelectInput(session, "bec_method", choices = methods)` with no `selected=` makes Shiny pick the first choice, ComBat. Every re-run of the comparison discarded the user's selection, and that is what reached `params$batch.correct.method`.

Fix: preserve the selection (`isolate()`d, so reading it does not make the reactive depend on it).

## Also

- `normalizeMethylation`'s `try()` result was checked for NULL but not `try-error`, so a failure assigned the error object to X and the panel went blank silently.
- `plot.outlierPCA` had the same NA-unsafe outlier flag `cleanX` just fixed.

## Verification

`dev/compute-comparer/` — a harness that drives the real upload UI with Playwright, captures the `params.RData` the app hands to `bin/pgxcreate_op.R`, and diffs it against a script-built one.

- `test_cleanX_parity.R` — 13 assertions, loads `realign_counts_to_X` **from the module**, so it fails if the fix is reverted (verified).
- `test_preprocess_sweep.R` — **45/45** across 9 fixtures × 5 option sets, covering methylomics, multi-omics, Olink NPX, metabolomics and four duplicate-bearing datasets.
- End-to-end, edgy app vs edgy script: params **40 fields + 13 preprocess options identical**; `createPGX` **5/5**; full `.pgx` **27 pass / 0 fail**.

Both tests run in seconds, no browser needed:

```
Rscript dev/compute-comparer/test_cleanX_parity.R
Rscript dev/compute-comparer/test_preprocess_sweep.R
```

## Behaviour change

Datasets with duplicated feature names now keep every feature through the preview, where before some were silently discarded. Duplicate handling proper still happens downstream in `createPGX` (`average.duplicated`).

## Not fixed here

`docs/FINDINGS.md` records a third bug left for a separate change: `input$upload_wizard` carries the step *title*, not `step_id`, so every wizard lock/unlock branch in `upload_server.R` is dead code and the wizard never locks.


zitoa:
@ESCRI11 looks ok - i'd remove all comments introduced by Claude. Also can you please send test data so i can test it?

zitoa (approved):

PR #1843, opened 2026-08-14
Build on debian:trixie (R 4.5 / BioC 3.22)

ESCRI11:
Companion to bigomics/playbase#511. **Both must land together** — see "The two pins" below.

## What this is

Builds Omics Playground on `debian:trixie` instead of `ubuntu:24.04`. The distro swap is not cosmetic: trixie ships R 4.5.0, and BiocManager hard-refuses BioC 3.18 on R 4.5, so R 4.3→4.5 and BioC 3.18→3.22 ride along unavoidably. That jump, not the `FROM` line, is the real risk.

The base image itself lives in playbase. This side is small: point at the trixie base, bump quarto, pin the package snapshot, and make CI able to build a branch at all.

## Status: validated end to end

Built from scratch on a Debian 13 box (`bigomics/playbase:trixie` → stage 1 → stage 2) and verified at runtime, not just built:

- Debian 13 · R 4.5.0 · BioC 3.22 · quarto 1.8.27 · typst · playbase 4.1.2 (edgy)
- app serves HTTP 200, titles as `Omics Playground 4`, no errors in log
- `example-data.pgx` loads: 18 samples, 7118 features, 5 real contrasts
- base graphics, a stored sample embedding, and a ggplot volcano over 7051 features all render (visually inspected)
- report export produces real PDFs through **both** engines — typst (posters) and LaTeX/tinytex (slides)

## ⚠️ Known issue this does NOT fix: report export breaks under AppArmor

On a Debian 13 host, `quarto render` inside a normally-confined container panics:

```
Deno has panicked ... failed to create UnixStream: Os { code: 13, kind: PermissionDenied }
```

No PDF is produced. **Both** report engines are affected. The only fix found is running the container with `apparmor=unconfined`; seccomp and the container runtime were both tested and ruled out. Note `quarto --version` still works, so a version-check smoke test reports healthy while render is broken.

This needs a deployment-side change (ShinyProxy container options) before trixie goes anywhere real. It was never seen in the earlier validation because that ran on WSL2, which has no AppArmor.

CI is unaffected — `setup-buildx-action` defaults to the `docker-container` driver, whose buildkitd is privileged and therefore unconfined. Local `make docker` on a Debian 13 machine **will** hit it, since all three Makefile targets use plain `docker build`.

## The two pins

`dev/Rprofile` here and `dev/rspm.R` in playbase must carry the **same** snapshot date. PPM's `latest` served lattice 0.23-1, which dropped the `parallel` export methylumi still imports; that cascades methylumi → lumi → wateRmelon, and wateRmelon is a playbase Import.

Pinning only one is **worse than pinning neither**: the base compiles correctly against lattice 0.22-9, then these stages upgrade lattice over it, and the build exits **0** with methylumi/lumi/wateRmelon installed but unloadable. `x %in% installed.packages()` returns TRUE for all three. Only `loadNamespace()` catches it.

Also worth knowing: PPM's source index advertised 0.22-9 while the binary channel shipped 0.23-1, so checking `src/contrib/PACKAGES` does not reveal this.

## CI changes

`docker-build.yml` hardcoded `ref: master` on checkout, so a `branch: <x>` dispatch built that branch's app code against **master's** Dockerfile and master's `dev/Rprofile` — on a trixie branch that silently reverts to the noble PPM URL, i.e. Ubuntu binaries on Debian. Checkout now follows the branch being built. Also:

- branch resolved once into `BUILD_BRANCH`, adding `github.head_ref` so PR builds stop passing `<PR#>/merge` to `git clone -b`
- `ARG BASE_IMAGE` before `FROM` in both Dockerfiles, so a trixie base can be used without retagging `:beta` out from under Docker Hub
- dropped `GITHUB_PAT` from `build-args` — unused (no matching `ARG`) and build-args land in image history; the BuildKit secret mount already delivers it
- `make docker.update` now passes `--secret`. Without it the private bigomics installs are skipped by their `if [ -n "$GITHUB_PAT" ]` guard **and the build still exits 0**

## Not done

- **The trixie playbase base is not on Docker Hub.** It exists only on the build host, so CI cannot do a full trixie build until someone publishes it.
- The AppArmor runtime issue above.
- `docker-base.yml` references `./docker/Dockerfile.base`, which does not exist in this repo — that workflow is dead. Left alone.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -9,6 +9,10 @@ on:
         description: 'Branch to build'
         required: true
         default: 'master'
+      base_image:
+        description: 'Base image to build FROM'
+        required: true
+        default: 'bigomics/omicsplayground:beta'
       update_playdata:
         description: 'Update playdata package'
         required: true
@@ -43,12 +47,20 @@ jobs:
     # runs-on: ubuntu-latest # free GH worker, not enough memory
     runs-on: self-hosted # run on dev-bigomics-frankfurt self-hosted runner
 
+    env:
+      # Branch to build: manual dispatch, else the PR source branch, else the
+      # pushed branch. Drives BOTH the checkout and the in-container git clone
+      # -- they must agree, or a branch build uses master's Dockerfile and
+      # master's dev/Rprofile while only the cloned app code follows the branch.
+      BUILD_BRANCH: ${{ github.event.inputs.branch || github.head_ref || github.ref_name }}
+      BASE_IMAGE: ${{ github.event.inputs.base_image || 'bigomics/omicsplayground:beta' }}
+
     steps:
       -
-        name: Checkout 
+        name: Checkout
         uses: actions/checkout@v3
         with:
-          ref: master
+          ref: ${{ env.BUILD_BRANCH }}
       -
         name: Get version
         id: get_version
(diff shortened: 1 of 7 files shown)


ivokwee:
Can this be merged? Also check https://github.com/bigomics/omicsplayground/pull/1870

PR #1846, opened 2026-08-16
perf: make the plot editor cheap (lazy content, per-session theme observers)

ivokwee:
Profiling the app under load (`Rprof`, R saturated at 100% CPU while switching boards) put a large share of tab-switch cost in board materialisation. The plot editor was a big part of it. This PR removes that cost without changing any behaviour.

## 1. Build only the editor variant that is used

`getEditorContent()` assigned all 18 `plot_type` variants and then `switch()`ed to return one — 17 full editor UIs built and discarded per call, 66 times.

`delayedAssign()` per variant, so each expression is untouched and only the selected one is forced.

**207ms → 81ms per call** (13.6s → 5.3s across the 66 modules).

## 2. Build the editor body only when the editor is opened

The rest was still built at materialisation and parked in a modal nobody had opened.

The modal *shell* is still emitted up front — the edit button is a plain Bootstrap trigger (`bigdash::modalTrigger`) targeting `#<ns>plotPopup2`, so the element must exist — but its body is now a `uiOutput`. A Bootstrap modal is `display:none` until opened, so Shiny suspends that output and the body is not built until the editor is opened.

Registering the renderer from UI-construction code needs `session$defineOutput()` (there is no `output` object there); with no reactive domain it falls back to building inline, unchanged.

Per plot module, inside a session:

| | before | after |
|---|---|---|
| DOM nodes | 306 | **9** |
| form inputs | 42 | **0** |
| build time | 81ms | **3ms** |

Across 66 modules: **~12,600 DOM nodes and ~700 bound inputs no longer created**, and ~5.3s → ~0.2s of R time at materialisation. Those inputs also stop being carried through Shiny's per-flush bookkeeping, which the profile showed at ~12% of R time (`manageHiddenOutputs`).

## 3. Theme observers installed once per session, not once per plot module

`plotmodule_theme_observer()` is the bigdash `editor_theme_observer` hook and ran per plot module: 9 observers each (8 colour keys + palette), all `ignoreInit = FALSE`, each issuing a separate `runjs()` per mapped input id.

Now installed once per browser session with a registry of target module sessions on `session$userData`; modules are seeded with the current theme when they register.

Measured over 70 modules:

| | before | after |
|---|---|---|
| observers installed | 630 | **9** |
| `runjs()` at materialisation | 840 | **70** |
| `runjs()` per colour change | 140 | **1** |

Verified a colour change still reaches all 70 modules.

## 4. Two correctness fixes found on the way

- **Colour store was a process-level singleton.** `reactiveValues()` registers an `onDestroy` handler on the domain current at creation, so the store built in the first browser session was already dead for the second: reload the page and the first write failed with `Can't access reactive ...; its module session has been destroyed`. It also leaked one user's theme into the next session. Now one store per session, anchored on `rootScope()`. *(This is PR #1845, included here as an ancestor — merge or close that one as superseded.)*

- **`dataview_plot_expression.R` read `input$scatter_color` raw**, while every other call site uses `get_editor_color(input, "scatter_color", "secondary")`. That input is a plot-editor input, so it is NULL whenever the editor is absent or not yet seeded, and `points.color[which(...)] <- NULL` fails with *"replacement has length zero"*. Matters more now that the editor is lazy.

- **The LLM warning fired at session start.** `observeEvent(input$enable_ai, ...)` had no `ignoreInit`, so it raised the data-exposure warning on the initial input value, before the user had been near the settings panel. Its own comment says "when they switch AI on".

## Verification

- All 19 `plot_type` values produce identical output after the `delayedAssign` change. *(Naive HTML comparison reports every type as differing — bslib generates random accordion ids, so the same call twice is not identical either; ids are normalised before comparing.)*
- Across 12 plot types, shell + rendered body carries **every element id** the eager version produced; the only addition is the `editor_body` placeholder.
- Theme push still reaches all 70 registered modules after the observer change.

## Not included

The remaining profile findings — caching the expensive plot computations that re-run on tab return (`glasso` 2.46s, `heatmaply` 2.70s, `seriate` 2.16s, ~40% of R time) and per-board rather than per-group UI insertion — are a separate PR.

@ESCRI11 — the piece worth your eye is `session$defineOutput()` in `getEditorContent()`. It is the only way I found to register a renderer from UI-construction code, given `PlotModuleServer()` never receives `plot_type`/`title`/`outputFunc` and so cannot rebuild the content itself. If you would rather this moved into bigdash (shell built by `PlotModuleUI`, body rendered by `PlotModuleServer` via a new hook), say so and I will do it that way instead.

### DIFF
--- a/components/board.dataview/R/dataview_plot_expression.R
+++ b/components/board.dataview/R/dataview_plot_expression.R
@@ -270,7 +270,13 @@ dataview_plot_expression_server <- function(id,
           df$samples <- factor(df$samples, levels = df$samples)
         }
 
-        points.color[which(points.color == "black")] <- input$scatter_color
+        ## Via get_editor_color(), as dataview_plot_boxplot.R and
+        ## dataview_plot_tsneplot.R do: `scatter_color` is a plot-editor input,
+        ## so reading it raw yields NULL whenever the editor is absent or has
+        ## not been seeded yet, and `x[i] <- NULL` fails with "replacement has
+        ## length zero". The helper falls back to the theme's `secondary`.
+        points.color[which(points.color == "black")] <-
+          get_editor_color(input, "scatter_color", "secondary")
 
         if (gp$use_ggprism) {
           ## --- ggplot2 + ggprism path (ungrouped) ---
--- a/components/board.user/R/appsettings_server.R
+++ b/components/board.user/R/appsettings_server.R
@@ -98,6 +98,12 @@ AppSettingsBoard <- function(id, auth, pgx) {
     })
 
     ## Warn the user when they switch AI on.
+    ##
+    ## ignoreInit = TRUE, or this fires when input$enable_ai receives its
+    ## initial value -- i.e. at session start, popping the LLM data-exposure
+    ## warning (or the "No LLM server available" error) over the app before
+    ## the user has been near the settings panel. The warning is about the act
+    ## of switching AI on, so only a real change should raise it.
     shiny::observeEvent(input$enable_ai, {
       model <- input$llm_reports
       if (isTRUE(input$enable_ai)) {
@@ -114,7 +120,7 @@ AppSettingsBoard <- function(id, auth, pgx) {
           # showCancelButton = TRUE
         )
       }
-    })
+    }, ignoreInit = TRUE)
     
     shiny::observeEvent(
       {
--- a/components/board.user/R/appsettings_ui.R
+++ b/components/board.user/R/appsettings_ui.R
@@ -189,7 +189,10 @@ AppSettingsUI <- function(id) {
       ),
 
       # Plot Colors #####
-      bslib::nav_panel(
+      ## DEVMODE-only: with the plot editor's body built lazily, theme changes
+      ## reach plot modules only through DOM-bound inputs, so a palette change
+      ## does not apply until that plot's editor has been opened (PR #1846).
+      if (isTRUE(opt$DEVMODE)) bslib::nav_panel(
         "Plot Colors",
         bslib::layout_columns(
           height = "calc(100vh - 85px)",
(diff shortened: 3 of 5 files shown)


ESCRI11 (changes requested):
main idea is good, however introduces functionality losses:

## 1

1. Start the app, load any dataset.
2. Go to App Settings → Default Palette, change default → muted.
3. Go to Clustering → PCA/tSNE (or Correlation → top-correlated barplot). Do not open its editor.
4. Expected: groups recoloured with the muted palette. Actual: still the default palette.
5. Now click the pencil on that plot. It jumps to muted the moment the editor renders — proving the value was never delivered, only re-derived on open.

## 2

1. Fresh session, load a dataset (board materialises → ct snapshot taken here).
2. App Settings → Primary colour: change red #f23451 → green.
3. Open Expression → Volcano. Up-regulated points are green. Correct so far.
4. Click the pencil on that same plot.
5. Expected: picker shows green. Actual: picker shows red — and binding sends red back, so the plot reverts to red.

## 3

1. Start the app on a deployment with ENABLE_AI on.
2. Expected (pre-PR behaviour): the "Using LLM might expose some of your data to external LLM servers" alert at session start. Actual: no alert, ever. Use any AI feature — still nothing.
3. To confirm it's the init suppression, not a broken alert: App Settings → Enable AI, toggle off, then on. The alert appears.


Number 3 might be relevant to fix for legal reasons. 1 & 2 i think its just a degradation on UX but not critical

ivokwee:
True for 1&2.. But I think the performance gains are more important. Hopefully we find a fix for such glitches. 

We can put back 3 if needed. No relation to plot editor. I Rwould localize it to Report & Obi.

ivokwee:
Can you confirm the improved app speed & reactivity?

ESCRI11:
hidden app settings plot palettes defaults under devmode for now to avoid 1&2 being problematic

ESCRI11 (approved):

PR #1848, opened 2026-08-16
WIP: perf: load one board per click instead of the whole module group

ivokwee:
## Why

Clicking a tab materialised its **entire module group**. Open Differential Expression and you also built and started Correlation, Biomarker and TimeSeries.

Measured first-click cost per group, instrumenting the two halves separately:

| group | total | UI insertion | server start |
|---|---|---|---|
| isect | 9.3s | **6.0s** | 3.3s |
| clustersamples | 6.6s | **4.8s** | 1.8s |
| drug | 5.7s | **3.9s** | 1.8s |

65–73% of it was building UI for boards the user had not asked for.

## What

`load_board()` loads one board. Both halves have to move together — a board server whose UI is not in the DOM loses the `updateSelectInput()` calls it makes while initialising — so the UI is inserted first, then that board's server starts, then the tab is latched (latched *before* running, so a board that errors does not re-insert its UI on every later click).

- `insertBigTabUI2()` gains `only =`, to insert a single tab's UI.
- Each group gains **`module_servers`**, its per-board servers keyed by tab name. `module_server()` is untouched, so the eager path still works.
- The 8 hand-written `if (input$nav %in% c("a-tab","b-tab",...) && loaded$x == 0)` blocks become `if (input$nav %in% names(MODULE.x$module_servers))`, so the tab list has one source of truth.
- `reset_group_tabs()` clears a group's latches wherever it is torn down on a datatype change.

**`diffexpr` and `enrich` are registered as UI-only** (`function(...) invisible(NULL)`). Their servers already start eagerly at dataset load — `env$diffexpr` is read by Enrichment and Compare, so it cannot wait for a click — but without an entry `load_board()` would skip the tab and their UI would never be inserted.

## Guard

The nav observer now waits for `PGX$name`. Clicking a sidebar item while the dashboard is still initialising would otherwise latch a board as loaded while its `insertUI()` found no target, and the board would never appear. I hit exactly that while testing.

## Verification

In the running app, clicking **only** `isect-tab`:

| | before click | after |
|---|---|---|
| `isect-tab` | 2 nodes | **2460** |
| `comp-tab` | 2 | **2** |
| `cmap-tab` | 2 | **2** |
| bound outputs | 180 | 198 |

Previously that single click built all three. Also verified every group's `module_servers` names match its `module_menu()` — that check caught the two missing entries above before they became silent no-loads.

## Relation to bigdash

This is the same mechanism as `bigTabsLazy()` in bigdash PR #37, done here in opg's own nav observer because #37 is not merged and installing it would drop the settings-scoping work. When #37 lands, this can be swapped for it — the per-tab registry maps onto it directly, and `preload = TRUE` is the equivalent of the UI-only entries.

@ESCRI11 — the part worth a careful look is the UI-only registration for `diffexpr`/`enrich`. It works, but it encodes "this board's server is started elsewhere" as a no-op function, which is easy to misread.

### DIFF
--- a/components/app/R/opg_server.R
+++ b/components/app/R/opg_server.R
@@ -205,6 +205,7 @@ opg_server <- function(input, output, session, PGX, env, auth, reload_pgxdir) {
             info("[SERVER:UI:1] calling Expression module")
             info("[SERVER:UI:1] calling DiffExprBoard module")
             bigdash.showMenuElement(session, "Expression")
+            ## preload expression board
             ExpressionBoard("diffexpr",
               pgx = PGX, labeltype = labeltype
             ) ->> env$diffexpr
@@ -220,6 +221,7 @@ opg_server <- function(input, output, session, PGX, env, auth, reload_pgxdir) {
             info("[SERVER:UI:1] calling GeneSets module")
             info("[SERVER:UI:1] calling EnrichmentBoard module")
             bigdash.showMenuElement(session, "GeneSets")
+            ## preload enrichment board
             EnrichmentBoard("enrich",
               pgx = PGX,
               selected_gxmethods = env$diffexpr$selected_gxmethods
--- a/components/app/R/opg_server.R
+++ b/components/app/R/opg_server.R
@@ -82,6 +82,7 @@ opg_server <- function(input, output, session, PGX, env, auth, reload_pgxdir) {
         })
         bigdash.hideMenuElement(session, "Clustering")
         loaded$clustering <- 0
+        reset_group_tabs(MODULE.clustering)
       }
       if (x == "Expression") {
         lapply(names(MODULE.expression$module_menu()), function(x) {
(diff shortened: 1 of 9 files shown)


ivokwee:
Follow-up once bigdash #37 (`bigTabsLazy()`) is merged and installed: most of this can be deleted.

- `load_board()`, the `loaded_tabs` environment and `reset_group_tabs()` become bigdash's job
- `insertBigTabUI2(only =)` is no longer needed
- the eight `if (input$nav %in% names(MODULE.x$module_servers))` blocks collapse into one `bigTabsLazy(list(...))` registration
- the `function(...) invisible(NULL)` entries for `diffexpr`/`enrich` become entries with `ui` and **no** `server` — `bigTabsLazy()` allows either half to be omitted, which says "UI only, server starts elsewhere" far more clearly than a no-op function

`module_servers` mostly survives as the registry, but each entry gains a `ui` thunk (today the UI comes from `module_ui2()`).

Two prerequisites: #37 has to merge, **and** the installed bigdash must also carry the sidebar/settings scoping from bigdash #34 — installing a build with only one of them regresses the other.

One open question for that pass: whether the `req(PGX$name)` guard is still needed, or whether `bigTabsLazy()` should handle the same DOM-timing case itself (a tab clicked before its shell exists gets latched while its `insertUI()` finds no target).

ivokwee:
Abandanned. Doing it differently with bigdash.

PR #1852, opened 2026-08-20
refactor(boards): migrate boardHeader() to OmicsBoardUI()/OmicsBoard()

ivokwee:
## Summary

- Migrates every remaining `board.*` package (21 packages, some with multiple sub-boards: `mofa`/`snf`/`mgsea`/`deepnet`/`lasagna`, `wgcna` consensus/preservation/multi, `user` appsettings/userprofile), plus `app_across` and `modules/SummaryBoard.R`, from the bare `boardHeader()` + `observeEvent(input$..._info, showModal(...))` pattern onto `OmicsBoardUI()`/`OmicsBoard()` — matching what `dataview`, `clustering`, `expression`, `wgcna` and all of `app_qsee` already use.
- This is prep work: `OmicsBoardUI` is the natural single injection point for wiring up bigdash's `bd_visibility_probe()`/`bd_is_visible()` later, so hidden boards' Plotly/iheatmapr plots can be purged app-wide the way `app_qsee` already does. That wiring is a deliberately separate follow-up, not part of this PR.
- Fixes a latent bug in `OmicsBoardUI()` itself (`components/ui/ui-boardHeader.R`): its `bslib::layout_columns(row_heights = list("auto", 1), ...)` silently assumed callers always pass exactly one object for the content row. `board.upload`'s original UI had three sibling elements instead of one; converting it as-is broke the header/content grid (a big blank gap above the visible content, no error). Wrapped `OmicsBoardUI`'s `...` in `div(...)` — a no-op for every existing single-child caller, but makes this class of bug impossible for any future board.

## Test plan

- [x] Every changed file parses (`Rscript -e "parse(...)"`) and was diff-reviewed line-by-line against the original `boardHeader()` call site to confirm only the header/info-modal block was touched.
- [x] Live-clicked through every board reachable with the bundled `example-data` (proteomics) dataset on `dev/run_app.R`: header renders (title, video-tutorial icon, current-dataset display), info modal opens with the correct text/video, no console errors.
- [x] Special cases verified individually: `board.admin`/`userprofile` (no `pgx`, pass `pgx = NULL`), `board.correlation` (iframe folded from modal title into `infotext`), `board.upload` (multi-child bug found, fixed centrally in `OmicsBoardUI`, re-verified).
- [x] WGCNA consensus/preservation/multi and the MOFA family boards need a multiomics dataset to reach in the UI — not available in this session, so those were verified via parse + diff only (identical structure to already-verified siblings in the same families).

### DIFF
--- a/components/app_across/R/across_server.R
+++ b/components/app_across/R/across_server.R
@@ -125,17 +125,11 @@ AcrossBoard <- function(id, pgx, pgx_dir = reactive(NULL),
 
     query_result <- reactiveVal(NULL)
 
+    OmicsBoard("board", pgx, title = "Across Datasets", infotext = infotext)
+
     ## ================================================================================
     ## ======================= OBSERVE FUNCTIONS ======================================
     ## ================================================================================
-
-    shiny::observeEvent(input$info, {
-      shiny::showModal(shiny::modalDialog(
-        title = shiny::HTML("<strong>Across Datasets</strong>"),
-        shiny::HTML(infotext),
-        easyClose = TRUE, size = "l"
-      ))
-    })
 
     shiny::observeEvent(input$zscore_info, {
       shiny::showModal(shiny::modalDialog(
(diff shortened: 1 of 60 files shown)


ESCRI11 (approved):

ivokwee:
Perhaps OmicsBoard and OmicsBoardUI shoudl eventually be migrated to bigdash. Let's see.

PR #1855, opened 2026-08-21
feat(boards): complete the bigTabsLazy migration

ivokwee:
Completes the migration of the board layer to `bigdash::bigTabsLazy()`, so board modules are registered and instantiated lazily instead of all up front.

Stacked on #1852 — review that one first; this PR's diff is the incremental work on top.

### What changes

- **Lazy module registration.** Boards register a factory rather than eagerly constructing their UI/server pair, so a board's module is only created when its tab is first opened.
- **Finish the `bigTabsLazy()` migration.** Remaining boards moved over to the lazy path.
- **One registry per menu group** instead of two parallel registries, removing the name/order coupling between them.
- **Group registries moved next to the app that reads them**, so each app owns its own registry rather than sourcing a shared central list.

Net effect is a smaller startup cost and ~600 fewer lines than the eager wiring it replaces.

### Notes

- The first commit is a checkpoint of in-flight state; the four that follow are the migration proper.
- Base should be retargeted to `edgy` once #1852 merges.


ivokwee:
after approval I will need to cleanup code.

ESCRI11:
Pushed 7bc529189: `EpigenomicsBoard` was still calling `OmicsBoard("board", ...)` after this branch changed the signature to `OmicsBoard(session, ...)` and dropped the internal `moduleServer`. `$` on a character vector is a hard error, not `NULL`, so the whole board module aborted at init — verified in a browser: Shiny session disconnected, empty dropdowns, spinners forever. One-word fix, board loads again.

ESCRI11 (approved):

PR #1857, opened 2026-08-21
fix: gate copilot document upload by extension and size

phisanti:
## Summary

The Copilot document panel accepted any file and listed whatever landed in `docs_dir`, while the agent only ingests `.pdf`/`.txt`/`.md`.  A user (Marteens) uploaded a large json. Agent could not read it and this lead to frustration. Then, converted to .txt, but file exceeded size (6MB, ~120K tokens), so the agent still could not read it. 

## Changes

- I have limited upload file types to what the agent can use (`.pdf`/`.txt`/`.md`)
- I have increased the max file size to 20MB (~5M tokens)

## Dependency

Requires **omicsagentovi >= 0.5.90** (pinned in `DESCRIPTION`). The gate allows 6 MB uploads because the backend cap is now 20 MB; against an older agent the backend still refuses at 500 KB and the silent-drop bug returns.

Merge [bigomics/omicsagentovi#<N>](https://github.com/bigomics/omicsagentovi/pull/3) first.


phisanti:
Please, @zitoa test this branch in your computer so that we can make sure we are align. I will not merge until you geen the green light.

ESCRI11 (approved):

PR #1858, opened 2026-08-23
feat(upload): gate PGX computation behind explicit Start confirmation

ivokwee:
## What
Refactors the upload wizard compute step:

- **Synchronous probe-type check** — `playbase::check_species_probetype()` now runs inside the compute step behind a blocking 'Checking your data...' alert, replacing the background `ExtendedTask` and the `probetype` reactive ('running'/'error' states) that was plumbed through `UploadBoard`.
- **Mismatch handling** — on failure the alert hints at alternative organisms (or valid metabolomics probe IDs), the upload is aborted and reset via a new `clear_upload()` helper (extracted from the `new_upload` observer).
- **Compute gated behind explicit confirmation** — the 'Crunching your data!' popup now has Start/Cancel buttons instead of a 60s timer auto-start; the processx job is dispatched only after the user clicks Start (`dispatch_computation` signal).
- **Wizard id renamed** `upload_wizard` → `upload_wizard_nav` to avoid collision with the reactive passed into the compute module.
- **Preview fix** — counts preview `table_data` is now `eventReactive(uploaded$counts.csv)` so it re-renders after a fresh upload or reset.

## Out of scope (left in working tree for separate PRs)
`app/R/ui.R` (`useShinyalert(force = TRUE)`), `00SourceAll.R` group-registry sources, `etc/OPTIONS`, `VERSION`.

## Testing
- Upload example/batch data, walk the wizard to the compute step: probe-check popup → Start / Cancel
- Mismatched probes for the selected organism → error alert + upload reset
- Methylomics upload → probe check skipped

### DIFF
--- a/components/board.upload/R/upload_module_computepgx.R
+++ b/components/board.upload/R/upload_module_computepgx.R
@@ -41,8 +41,8 @@ upload_module_computepgx_server <- function(
   upload_gset_methods,
   process_counter,
   reset_upload_text_input,
-  probetype,
-  recompute_pgx = NULL
+  recompute_pgx = NULL,
+  clear_upload
 ) {
   shiny::moduleServer(
     id,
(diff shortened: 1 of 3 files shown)


ivokwee:
Discarded.

PR #1859, opened 2026-08-24
fix: eliminate 8-10s lag on first dropdown action in Data Library

ivokwee:
## Problem

Clicking the ellipsis (⋮) button next to a dataset in the Data Library, then selecting an action like "Change name" or "Delete dataset", caused an **8-10 second delay** before the confirmation modal appeared. Subsequent clicks on the same action were instant.

## Root cause

A single line in `ui.R`:

```r
visnetwork = visNetwork::visNetworkOutput("a", height = "0px"),
```

This was a **named argument** inside `shiny::tagList()`. In R, `tagList()` treats named arguments as HTML attributes, **not children**. Everything after this line — including `shinyalert::useShinyalert(force = TRUE)` — was silently dropped from the DOM.

The `shinyalert` JavaScript dependencies (sweetalert, swalservice, shinyalert binding) were never loaded at page start. The first `shinyalert::shinyalert()` call triggered a lazy `insertUI("head", ...)` to inject the scripts, forcing the browser to download and execute them before the modal could appear — causing the 8-10 second delay.

## Solution

Three changes:

1. **`ui.R`** — Removed the orphaned `visNetworkOutput("a")` (no server-side handler existed). Wrapped the remaining dependency loaders (`shinyjs`, `shinyalert`, `waiter`, `sever`, `bigLoaders`, `shinybrowser`) in a proper unnamed `shiny::tagList()` so they are actually rendered in the DOM. Added a comment warning about the named-argument trap.

2. **`server.R`** — Added `session$userData$.shinyalert_added <- TRUE` at session startup so `shinyalert::shinyalert()` skips its redundant lazy `insertUI` call on first use.

3. **`loading_table_datasets.R`** — Removed a nested `observeEvent` from inside the `pgxTable_DT` reactive (observer in wrong scope). Cached the `getFilteredPGXINFO()` call in the changename handler to avoid double evaluation.

### DIFF
--- a/components/app/R/server.R
+++ b/components/app/R/server.R
@@ -18,6 +18,10 @@ app_server <- function(input, output, session) {
   ## Initialise the global colour theme (in-session only)
   init_color_theme()
 
+  ## Mark shinyalert dependencies as already loaded so the first call
+  ## skips lazy insertUI (which causes 8-10s delay while JS downloads).
+  session$userData$.shinyalert_added <- TRUE
+
   info("[SERVER] getwd = ", normalizePath(getwd()))
   info("[SERVER] SESSION = ", session$token)
 
--- a/components/app/R/ui.R
+++ b/components/app/R/ui.R
@@ -53,12 +53,22 @@ app_ui <- function(x) {
       shiny::tags$head(shiny::tags$script(src = "static/shared-badges.js")),
       shiny::tags$head(shiny::tags$link(rel = "stylesheet", href = "custom/styles.min.css")),
       shiny::tags$head(shiny::tags$link(rel = "shortcut icon", href = "custom/favicon.ico")),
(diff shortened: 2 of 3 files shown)


ivokwee:
@ESCRI11  why was this inserted anyway? if not needed let's remove. very weird it got there

`         visnetwork = visNetwork::visNetworkOutput("a", height = "0px"),`

ESCRI11 (approved):
ok two details: the bug this PR appears to fix, i cant reproduce either locally or in the edgy deployed versions, so not sure what are we trying to solve here at all

the weird line was yet another hack i had to do two years ago so visnetwork worked when i did the v1.0 of the progressive UI load, seems after two years and some changes its not needed so it can go

https://github.com/bigomics/omicsplayground/commit/49955301808c3c1aca6c26c16528027360387451

PR #1863, opened 2026-08-24
Align pages exactly to the browser height

ivokwee:
Pages rendered at content height inside 100%-height containers, so boards
and full-page views overflowed the window instead of fitting it.

## Fill chain

Chain the fill through the board content row, the tabset wrapper and the
active pane, so nested `layout_columns()` resolve against a definite height.

Data Library, AI Studio and Obi are mounted as `bslib::nav_panel()` and sit
outside `.big-tab`, so no board rule reached them. Their wrappers are marked
`.fullheight-page`, which has its own rules anchored with **child**
combinators — `navset_hidden()`/`navset_underline()` render the same
`.tabbable > .tab-content > .tab-pane` markup, and descendant selectors
forced `display:flex` onto the seven navsets nested in AI Studio's cards,
stacking their panels.

Also zeroes bslib's 24px `.bslib-mb-spacing` margin on full-height grids.

## `row_heights`

`c("auto", 1)` is a **silent no-op**: `c()` coerces the `1` to `"1"`, bslib
emits the invalid `auto 1`, CSS drops the declaration and both rows stretch.
Fixed at all 16 sites by using `list("auto", 1)`.

Added `row_heights` to 31 `layout_columns()` holding a `bs_alert()`, which
otherwise split free space evenly and padded out the alert row. Fractional
ratios are now whole numbers — bslib rounds `fr` values to integers, so
`list(1.33, 1)` was rendering `1fr 1fr`.

## Data Library

`fillRow()` hardcodes `height:100%` with an absolutely-positioned inner, so
it has no content height; as a flex child it took half the page. Replaced
with a plain flex row. Table and t-SNE go `100vh` → `100%`.

## Polish

17px of dead space removed above the board title; padding on the board
content row so the bottom card's border is not clipped by the overflow clip;
46 dead height locals dropped.

## Screenshots

Both pairs captured at the same 1520x757 viewport, before and after the change.

**Differential expression** (board) — the content row overflowed by 488px, so the
table row below the plots was cut off and the board grew its own scrollbar.

| Before | After |
| --- | --- |
| _drop `board-differential-expression-BEFORE.jpg` here_ | _drop `board-differential-expression-AFTER.jpg` here_ |

**Data Library** — the page overflowed by 221px past the window.

| Before | After |
| --- | --- |
| _drop `data-library-BEFORE.jpg` here_ | _drop `data-library-AFTER.jpg` here_ |

## Verification

Boards and pages were measured in the browser (`scrollHeight - clientHeight`
per container) rather than eyeballed: zero overflow on every board checked,
all four Data Library tabs, AI Studio and Obi, with nested navsets confirmed
still showing one panel at a time.

Needs a look on a real screen before un-drafting.

https://claude.ai/code/session_01N87p7YNKPgYXMdCUGY7nkr


ivokwee:
Differential expression BEFORE:
board-differential-expression-BEFORE

Differential expression AFTER:
board-differential-expression-AFTER

ivokwee:
Data Library BEFORE:
data-library-BEFORE

Data Library AFTER:
data-library-AFTER

ivokwee:
I think all these tweeks should move to bigdash some time again

ESCRI11 (approved):

PR #1873, opened 2026-08-31
Obi display switches, bulk history delete, and UI fixes

ivokwee:
Changes in Obi (Copilot) plus a round of UI/layout fixes across Studio, Upload, Loading and Settings.

## Obi / Copilot

**Display switches in the chat settings panel** — two new checkboxes, both display-only:

- **Show reasoning trace** (default off). Reasoning is *always* requested from the model (both tiers run `reasoning_effort = "medium"`); the switch only decides whether the trace is streamed into the answer bubble, wrapped in grey italic brackets ahead of the answer. It is never written to saved transcripts, so a live run and its restored replay stay identical. Seeded by `options(copilot.show_thinking = ...)`.
- **Show tool calls** (default on). Withholds the `on_tool_request` callback, which only draws the inline marker — tool dispatch, argument normalization and run-state tracking are untouched. Seeded by `options(copilot.show_tools = ...)`.

Both are read by the run controller at dispatch time, so toggling takes effect on the next message and never interrupts a run in flight. The `show_thinking` arguments are passed via `do.call` behind a capability probe (`.stream_supports_show_thinking`) so an older `omicsagentovi` without those parameters still works.

**History: Delete all** — a "Delete all" button next to "Delete selected", behind a confirmation modal that names the count. `on_delete` now carries a character vector instead of a single id; the orchestrator loops (`ovi_session_delete` is per-session), collects failures into one notification instead of one per session, and refreshes the table either way. The button row is pinned with `flex: 0 0 auto` so it isn't squashed to zero by the table above it.

**Text/labels** — the "Documents" tab is now "Sources", with a short explainer that Obi already knows the current dataset and this is for extra context. Mascot background bumped to 120px.

## UI fixes

- **Studio**: menu title "AI Studio" → "Studio"; "Infographic" → "Infographics" throughout (UI, settings panel and the `show_infographics` observer, which was previously wired to a button id that no longer matched); headings normalized ("AI reports" → "AI Reports", "Infographics" → "AI Infographics").
- **Settings**: `navset_pill_list` extracted to a `panel` variable, `header_margin = "0px"`, App-settings column widths `c(5,7)` → `c(4,8)`, and the theme panel's colour pickers regrouped into an "Accent Colors" card. New `#app_settings-tabs1` SCSS rule bumps the left pill list to 1.05rem — via `> li > a` because bslib 0.12 / shiny 1.14 still emit BS3 tab markup with no `.nav-link` class on the anchor.
- **Upload / Loading**: dead duplicate `div()` removed from `LoadingUI` (only the second was ever returned), plus margin/spacing fixes on the upload wizard, normalization panel and comparisons buttons.
- **Global**: base font 14px → 15px, `.nav-link` set to 1rem; `styles.min.css` rebuilt.
- `VERSION` → `5.0-rc1+edgy260830`.

## Test plan

- [ ] Obi chat: toggle both switches, confirm reasoning brackets appear/disappear and tool markers hide while tools still run
- [ ] Obi history: "Delete all" confirmation, and "Delete selected" still works
- [ ] Studio: the Infographics button selects the Infographics panel
- [ ] Settings, Upload and Loading pages for layout regressions

https://claude.ai/code/session_01AGfWdqoS99f7YrC8RrzqFj


phisanti:
Finally, manage to reproduce. Please link this to > `ellmer 0.4.2`

phisanti (approved):

PR #1876, opened 2026-09-01
fix(biomarker): explain why Compute is greyed out

ESCRI11:
On **Find Biomarkers**, the Compute button is disabled when the sample filter leaves the prediction target with a single level. Nothing said so, so the button just looked broken. This renders the reason under the button.

| Filter collapses the target | Valid target |
|---|---|
|  |  |

### DIFF
--- a/components/board.biomarker/R/biomarker_server.R
+++ b/components/board.biomarker/R/biomarker_server.R
@@ -90,6 +90,24 @@ BiomarkerBoard <- function(id, pgx) {
       }
     )
 
+    ## explain why the run button above is greyed out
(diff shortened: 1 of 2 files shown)


phisanti (commented):
maybe not red text. Red == bug? Greyed?

phisanti (approved):
Otherwise, approved

ESCRI11:
Switched to grey — you're right, red reads as "bug". Using `#555`, dark enough to stay readable on the light settings panel.

Compute disabled with grey explanation

c35f0380c

PR #1877, opened 2026-09-01
fix(ui): equal rows in nested layout_columns (squashed bottom row)

ESCRI11:
Nested `layout_columns()` grids that don't pass `row_heights` were rendering their first row content-sized and squashing the second: **454px / 157px** in a 635px board.

`row_heights=` is emitted by bslib as the custom property `--bslib-grid--row-heights` and read back as `grid-auto-rows: var(--bslib-grid--row-heights)`. Custom properties inherit, so every nested grid without its own `row_heights` adopted its ancestor's — and the ancestor is the board's outer grid carrying `list("auto", 1)` for the `bs_alert`. One rule supplies the default those grids never asked for; bslib writes an explicit `row_heights` inline, so it still wins.

```scss
.big-tab .bslib-grid {
    --bslib-grid--row-heights: 1fr;
}
```

Measured (1600x900): WGCNA Modules/Eigengenes/Enrichment, Compare Foldchange + Gene Correlation, Similar Experiments FC correlation all go from skewed to equal rows, no board overflow.

### WGCNA > Modules

| before | after |
|---|---|
|  |  |

### WGCNA > Enrichment

| before | after |
|---|---|
|  |  |

Notes:
- `components/assets/styles.min.css` is rebuilt with `Rscript dev/sass.R`; it also picks up `_obi.scss` changes from `fffe2014d` (already merged on edgy) that were never rebuilt: the Obi workspace/history fill-chain rules and the mascot shrink.
- One extra line: `dev/board.launch/global.R` had `APPDIR` pointing at `components/app/R`, so `make board.example` crashed on `utils/utils.R`. Needed it to reproduce.


phisanti (approved):
lgtm

PR #1879, opened 2026-09-01
fix(home): stop the launcher overflowing on 14" laptops

ESCRI11:
The Home screen was a fixed 994px tall regardless of viewport, so it overflowed on every 14" laptop size. The mascot alone cost 260px of `padding-top`.

- Mascot sized off `min(260px, 20vh)` instead of a flat 260px.
- Launcher page is a `min-height: 100vh` flex column, so the credits carousel uses `margin-top: auto` instead of a fixed 70px margin. This also fixes the tinted background stopping dead at `100vh` and leaving a white band under the tiles.
- Side padding 12% → 4%, dashboards grid to `auto-fit, minmax(240px, 1fr)` — all four tiles on one row instead of wrapping 3 + 1.

| viewport | content height before | after |
|---|---|---|
| 1280×720 | 994px (274 over) | 808px (88 over) |
| 1366×768 | 994px (226 over) | 797px (29 over) |
| 1512×945 | 994px (49 over) | 945px — fits |

### 1280×720

| before | after |
|---|---|
|  |  |

### 1512×945

| before | after |
|---|---|
|  |  |

Screenshots live on the `assets/better-home-screenshots` branch.

### Not fixed here

- 1280×720 and 1366×768 still overflow slightly, and page height oscillates 36px as the carousel rotates between its short and long slides. Halving the section margins (40 → 20px) closes it.
- `.app-tile-wide` is `height: auto; max-height: 140px`, so tiles are ~3px vertically misaligned.
- 12px horizontal overflow from `navset_pill_list`'s row gutters vs the parent's `p-0` — pre-existing, affects every board.


phisanti (approved):
Only css changes, correct.

PR #1880, opened 2026-09-01
fix(copilot): PNG download of plots produced no file

ESCRI11:
## Description

Downloading a Copilot plot as PNG did nothing. The browser reported a failed download, no error appeared anywhere in the UI, and it affected every plot kind except ggplot — which is most of them, since Copilot plots come from omicsplots and are mostly plotly.

Two separate causes:

- **plotly** went through `bigdash::plotlyExport`, which only really tries kaleido — its own webshot2 fallback is dead code sitting behind a hardcoded `export.ok <- TRUE`. Without the Python venv that `dev/install_extra.R` builds, it writes no file at all, and Shiny 404s the download.
- **iheatmapr** called `iheatmapr::save_iheatmap`, which is S4-dispatched on `"Iheatmap"`. But `omicsplots::pgx.plot_heatmap` already hands back the converted `"iheatmapr"` htmlwidget, so dispatch could never match. Broken in every environment, kaleido or not.

Both now fall back to one small `.widget_png` helper — save the widget to a temp HTML, screenshot it with `webshot2` — at the same 4x upscale the ggplot and kaleido branches use, with the plotly.js modebar hidden so it doesn't end up in the image. kaleido is still tried first for plotly when it's available, since it renders faster and sharper than a browser screenshot.

Also: the test file's `.board_dir` still pointed at `components/board.copilot/R`, which no longer exists, so the whole file errored at source time and its 30 assertions never ran. Repointed at `app_copilot`; they pass.

## Related Issue

None.

## Example

Ask Copilot for a volcano plot or a heatmap, then download icon → Format → PNG.

Tests drive the real `downloadHandler` through `shiny::testServer` and assert the returned file is a PNG at the requested resolution:

```
Rscript -e 'testthat::test_dir("tests/testthat", filter="copilot-evidence-server")'
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 36 ]
```

Reverting either branch of the fix turns those red.

## Not in this PR

Same kaleido root cause breaks PNG, PDF **and** SVG download for every `plotlib="plotly"` board plot (178 call sites) via `bigdash::plotModule`. The real fix is deleting that `export.ok <- TRUE` line in bigdash so its existing fallback runs. Two smaller ones also found while looking: `plotlib = "baseplot"` in `components/modules/UsersMapModule.R:104` isn't a value bigdash knows (should be `"base"`), and the ggplot PDF path calls `sysfonts::font_add_google("lato")` unguarded — a bigdash *Suggests* that no DESCRIPTION or `docker/packages.list` installs, and which needs network access at download time.

https://claude.ai/code/session_01QpqvooB4N7Ztpv8Q8sjcnt

### DIFF
--- a/components/app_copilot/R/CopilotEvidenceServer.R
+++ b/components/app_copilot/R/CopilotEvidenceServer.R
@@ -155,16 +155,41 @@ CopilotEvidenceServer <- function(id, local_pgx = NULL) {
     # ---- Download handler ----
     # Format choice depends on artifact kind:
     #   ggplot    -> PNG (only; ggsave at width x height inches, 288 dpi = 4x)
-    #   plotly    -> HTML (saveWidget) or PNG (plotlyExport: kaleido -> webshot2)
-    #   iheatmapr -> HTML (saveWidget) or PNG (iheatmapr::save_iheatmap, webshot)
-    # plotlyExport is defined in bigdash R/plot-module.R and is reused
-    # here so we don't fork the kaleido/webshot fallback chain.
+    #   plotly    -> HTML (saveWidget) or PNG (plotlyExport, else .widget_png)
+    #   iheatmapr -> HTML (saveWidget) or PNG (.widget_png)
+    # bigdash::plotlyExport (R/plot-module.R) is tried first because kaleido
+    # renders faster and sharper than a browser screenshot when it is there.
     .default_fmt <- function(kind) {
       switch(kind, ggplot = "png", plotly = "html", iheatmapr = "html", "html")
     }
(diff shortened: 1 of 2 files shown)


phisanti (approved):
lgtm, I would only shorten a bit the unhinged claude doc-string (` # Screenshot an htmlwidget to PNG via...`. It is 10 LOC of comment + 3 internal LOC docstring for 7LOC code

ESCRI11:
true, slopus 5 is a little bit too much sometimes; shortening comment and merging

PR #1882, opened 2026-09-01
feat: enable  mini apps (devmode only)

ivokwee:
## Summary
- Wire IDconvert and SmartPrism up to the `launchModule()`/`nav_insert()` pattern already used for Qsee/Across, so their Home page tiles actually launch the module on click instead of doing nothing (IDconvert/Prism servers were never attached) or selecting a static pre-rendered panel.
- Drop the now-redundant static `nav_panel_hidden()` blocks for IDconvert/Prism from `ui.R`.
- Add `padding = 0` to Prism's and MOFA's `bslib::page_fillable()` roots (IDconvert/Qsee already had this) so bslib's default 1rem page padding doesn't render as a thick white border around the module.
- Fix a crash-on-launch in `prism_server.R`: `get_plotcode()`'s eager `observe()` fired before the client sent back `chartbot_user_input`, so `isolate(input$chartbot_user_input) == ""` compared `NULL == ""`, producing `logical(0)` and an `if (argument is of length zero)` error that killed the whole session. This was latent since `prism_server()` was never actually called before this PR.

## Test plan
- [x] Reproduced the crash pre-fix by launching SmartPrism from Home (session disconnected with "Woops! You have been disconnected")
- [x] Verified the fix with `shiny::testServer(prism_server, ...)`: initial reactive flush and an empty `chartbot_user_input` both complete without error
- [x] Manual click-through of IDconvert and SmartPrism tiles from Home in a running app session (DEVMODE=TRUE)

## NOTE
This need first playbase PR https://github.com/bigomics/playbase/pull/504  (bridge-ortholog) to be merged.


ESCRI11 (changes requested):
@ivokwee looks good. please move target to devel instead of edgy + solve merge conflicts. after feel free to merge

ESCRI11 (approved):

PR #1884, opened 2026-09-02
Gate DataView Settings tab behind DEVMODE

ESCRI11:
as discussed via slack:

The Settings tab is half-finished (the second card has no server binding), so hide it outside DEVMODE for now

### DIFF
--- a/components/board.dataview/R/dataview_ui.R
+++ b/components/board.dataview/R/dataview_ui.R
@@ -354,8 +354,10 @@ DataViewUI <- function(id) {
     )
   )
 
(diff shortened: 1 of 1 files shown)


ivokwee (approved):
Small, self-contained change: wraps the half-finished DataView Settings tab (unbound `settings2` card) behind `opt$DEVMODE`. Verified `shiny::tabsetPanel()` silently drops a NULL child, and the only other "Settings" reference in dataview_server.R is an unused lookup-table key when the tab doesn't exist. LGTM.

PR #1889, opened 2026-09-07
Fix/user dir

phisanti:
## Summary
Marteen mentioined that agents can find papers, docs and sessions but no .pgx objects.

## Changes
manage_pgx list forwards to omicspgxmcp, which resolves the directory from a process-global cache seeded once per Agent construction
a host scoping data_dir per user over one R process had the newest Agent own that global, leaking other sessions' dataset names

Depends on: https://github.com/bigomics/omicsagentovi/pull/6

### DIFF
--- a/components/app_copilot/R/CopilotBoardServer.R
+++ b/components/app_copilot/R/CopilotBoardServer.R
@@ -11,12 +11,14 @@
 #'
 #' @param id         Module namespace id.
 #' @param pgx        Global PGX reactiveValues shared across all boards.
-#' @param pgx_dir    Path to directory containing .pgx dataset files.
+#' @param pgx_dir    Root of the PGX data tree. Used only to derive the
+#'   per-user base directory; Copilot itself never reads datasets from this
+#'   level (see the path-resolution block below).
 #' @param auth       Auth module reactiveValues (must expose `$user_dir` and
-#'   `$email`). Copilot's `chat_dir` and `docs_dir` are always scoped to the
-#'   per-user folder `<pgx_dir>/<email>` when an email is known, independent
-#'   of ENABLE_USERDIR (see the path-resolution block below); with no email
-#'   it falls back to `auth$user_dir`.
+#'   `$email`). Copilot's `data_dir`, `chat_dir` and `docs_dir` are all scoped
+#'   to the per-user folder `<pgx_dir>/<email>` when an email is known,
+#'   independent of ENABLE_USERDIR (see the path-resolution block below);
+#'   with no email it falls back to `auth$user_dir`.
 #' @param maxturns   Maximum user turns per session.
 #' @param tiers      Character vector of tier identifiers (first = default).
 #'   Defaults to `COPILOT_TIERS` from `copilot_options.R`.
@@ -42,8 +44,10 @@ CopilotBoardServer <- function(
     # change, and the SessionStore + downstream controllers expect plain
     # character paths, not reactives.
     #
-    # Copilot chats and uploaded docs are personal, so they always live in
-    # the per-user folder <pgx_dir>/<email> when an email is known —
+    # `user_dir` is the single base for everything Copilot touches: the
+    # datasets it lists and loads, plus the chats and uploaded docs it writes
+    # alongside them. Datasets, chats and docs are all personal, so they live
+    # in the per-user folder <pgx_dir>/<email> when an email is known —
     # independent of ENABLE_USERDIR. That flag only governs *dataset*
     # storage; when it is off it collapses auth$user_dir to the shared
     # PGX.DIR, which would otherwise scatter every user's chats/docs into one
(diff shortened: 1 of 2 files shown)


components/app_copilot/R/CopilotBoardServer.R
```diff
@@ -183,11 +195,11 @@ CopilotBoardServer <- function(
                            session          = session,
                            evidence_api     = list(append_artifact = evidence$append_artifact),
                            docs_dir         = docs_dir,
-                           data_dir         = pgx_dir,
+                           data_dir         = user_dir,
```
ESCRI11, line 198: Datasets now go through the `<pgx_dir>/<email>` override, which is derived unconditionally from the email — but that folder only holds .pgx files when `ENABLE_USERDIR` is TRUE. In a deployment with `ENABLE_USERDIR = FALSE`, AuthenticationModule collapses `auth$user_dir` back to `PGX.DIR` and every dataset is written directly into `PGX.DIR` (see AuthenticationModule.R:399/478/679/918). Same for `AuthenticationModuleApacheCookie`, which sets `user_dir = PGX.DIR` while `email` is non-empty. In those modes Copilot now scans `PGX.DIR/<email>` — a directory that auth never creates (Copilot's own `dir.create(chat_dir)` makes it, empty) — so `CopilotDatasetsServer` shows "No datasets found" and `manage_pgx list` returns nothing, where before the change it listed the real datasets.

Scoping the *dataset* dir is only correct when the per-email layout is actually in use; the chats/docs override was safe precisely because it doesn't need pre-existing content. Suggest gating the dataset base on `opt$ENABLE_USERDIR` (i.e. use `auth$user_dir` for `data_dir`, keeping the email override only for `chat_dir`/`docs_dir`), or at least falling back to `auth$user_dir` when the derived folder contains no .pgx files.

Related: the new comment claims "scoping here hides nothing that was previously reachable" — that holds for the userdir layout, but not for the two configurations above.

phisanti:
Good catch — confirmed on all three points, this was a real regression in `ENABLE_USERDIR = FALSE` and apache-cookie deployments. Fixed in 6acc3b27d.

The mistake was re-deriving a value auth had already computed correctly. `auth$user_dir` *is* the ENABLE_USERDIR-aware answer to "where do this user's datasets live" — it is what `pgx_save_target()` writes to (`components/app/R/server.R:846`) and what the LoadingBoard reads (`board.loading/R/loading_server.R:160`). Under `ENABLE_USERDIR = TRUE` it already equals `PGX.DIR/<email>`, so the override was redundant there and wrong everywhere else. I carried it over from the chats/docs block without checking that its rationale transfers — it does not, for exactly the reason you give: chats/docs need no pre-existing content, datasets do.

Worth noting the failure was silent rather than loud: `dir.create(chat_dir, recursive = TRUE)` creates `PGX.DIR/<email>/` as a side effect, so the scan hit an existing empty directory and reported "No datasets found".

The fix splits the two bases:

- `user_dir` (datasets) — `auth$user_dir` verbatim, no derivation
- `personal_dir` (chats, docs) — keeps the per-email override, i.e. back to its pre-PR behaviour

Resolution across every auth path:

| config | `data_dir` | `chats` | datasets truly at |
|---|---|---|---|
| `ENABLE_USERDIR=TRUE` | `/data/u@x.com` | `/data/u@x.com/chats` | `/data/u@x.com` ✓ |
| `ENABLE_USERDIR=FALSE` | `/data` | `/data/u@x.com/chats` | `/data` ✓ |
| ApacheCookie | `/data` | `/data/u@x.com/chats` | `/data` ✓ |
| anonymous | `/data` | `/data/chats` | `/data` ✓ |

I skipped the "fall back when the folder holds no .pgx files" alternative deliberately — it cannot distinguish "wrong directory" from "user has no datasets yet", so a genuinely empty new account would silently fall back to the shared root, which is the disclosure shape the companion PR is trying to close. The `ENABLE_USERDIR` gate is exact; the emptiness test is a guess.

Also dropped the "hides nothing that was previously reachable" claim — you are right that it only held for the userdir layout.

ESCRI11 (approved):

ESCRI11:
@phisanti heads-up on **merge order** — this one needs `bigomics/omicsagentovi#6` to land first.

`data_dir` used to be a process-wide constant and is now per-user, but `manage_pgx(action="list")` doesn't read the per-session bindings — it reads a process-global:

- `omicsagentovi::Agent()` calls `omicspgxmcp::set_data_dir(bindings@data_dir)` at construction
- `omicspgxmcp/R/pgx-env.R:99` writes `.pgxmcp_env$data_dir` — one variable per R process, and it takes priority over the `getOption("omicspgxmcp.data_dir")` fallback we set in `global.R:64`
- `tool-manage_pgx.R:103` (`.pgx_listing()`) reads it at listing time

On the shipped defaults that's a single `runApp` process (`docker/Dockerfile:99`) with `AUTHENTICATION = password` and `ENABLE_USERDIR = TRUE`. Alice opens Copilot → global becomes `/data/alice@…`; Bob opens Copilot on the same container → global becomes `/data/bob@…`; Alice then asks "which datasets do I have?" and gets **Bob's** dataset names, dates and `datasets-info.csv` metadata. `load` is unaffected (it uses `bindings@data_dir` per call), so it's name/metadata disclosure rather than content — but pre-PR the global was `PGX.DIR` for everyone, so this is new behaviour introduced by the per-user scoping.

omicsagentovi#6 fixes it by re-pinning the global before each listing forward (0.5.95). Since `DESCRIPTION` currently declares `omicsagentovi (>= 0.5.90)` and `copilot_packages_ok()` never version-checks that Suggests floor at runtime, merging this PR alone would leave the leak live on 0.5.90–0.5.94.

So: merge omicsagentovi#6 → bump `DESCRIPTION` to `omicsagentovi (>= 0.5.95)` here → merge this.

Rest of the change reviewed and looks correct — `auth$user_dir` is the directory datasets are actually written to and read from in every auth mode, and the `personal_dir` split keeps chats/docs exactly where they were.

phisanti:
Version bumped and ready to merge if you give the last ok.

ESCRI11:
green light!

PR #1895, opened 2026-09-14
fix(ui): reset bar pushes plot editor below the fold

ESCRI11:
The editor's reset bar picked up `.big-tab .bslib-grid.html-fill-item { height:100% }` (from #1863), taking the full modal height. `fill = FALSE` opts it out.


phisanti (approved):
lgtm

PR #1896, opened 2026-09-14
fix: static clustering heatmap errors until the editor is opened

ESCRI11:
Static heatmap failed with `invalid 'x' type in 'x && y'`: editor inputs are NULL until the lazily built editor (dd098509b) is opened. Defaults them in the render.

| Before | After |
|---|---|
| before | after |


phisanti (approved):
LGTM

PR #1897, opened 2026-09-16
Fix/ai activation

phisanti:
Reported from the free deployment: the "Enable AI" switch is greyed out and
unresponsive. Initially looked tier-specific — it isn't.

## Cause

`AI_PROVIDER_LOCKED` fed `ai_base_locked()`, which gated three controls:

    ai_base_locked() = (AI_PROVIDER_LOCKED && !auth$ADMIN) || !ENABLE_AI
      -> disable("enable_ai")       <- collateral
      -> disable("ai_share_data")   <- collateral
      -> disable("ai_provider")     <- the only intended one

`free.playground.bigomics.ch` sets `AI_PROVIDER_LOCKED = TRUE` purely to stop
users picking a provider — its own OPTIONS comment reads "TRUE = users can't
pick their own provider". It also froze the AI on/off switch and the
data-sharing consent for every non-admin.

Confirmed live via DOM read on that deployment: all seven AI controls carried
`disabled=true`; `enable_ai` additionally had `pointer-events: none`,
`opacity: 0.5` and `checked=true` — AI forced on with no opt-out, despite the
app warning that LLM use may expose data to external servers. Reproduces for
free, pro and enterprise alike; only ADMIN escaped, which is why it went
unnoticed.

The flag is also redundant there: `ai_byok_allowed()` already returns FALSE
for free/pro, so the provider was already pinned to `bigomics` and greyed.

## Change

Three gates, now kept apart:

| gate | governs |
|---|---|
| `ai_base_locked()` = `!ENABLE_AI` | everything, incl. `enable_ai` + consent |
| `ai_provider_pinned()` = `AI_PROVIDER_LOCKED && !ADMIN` | provider dropdown only |
| `!byok` | provider dropdown only |

A deployment that wants AI off uses `ENABLE_AI=FALSE` — already the correct lever.

## Verification

Probe under the exact free-deploy config (`ENABLE_AI=TRUE`,
`AI_PROVIDER_LOCKED=TRUE`, `providers=bigomics`):

                enable_ai  ai_provider  ai_share_data  llm_*
    before free    off         off           off        off
    after  free    ON          off           ON         off
    after  pro     ON          off           ON         off
    after  ent.    ON          off           ON         off

Provider pin still works; only the two controls it was never meant to touch
came back. `test-appsettings-ai.R`: 33 tests / 111 assertions, 0 failures.

Three consent tests encoded the old coupling and were re-pointed at the
licence gate; added a regression guard for the free-deploy config.

## Deploy note

This ships with the next devel image. **The free deployment is fixed today by
removing `AI_PROVIDER_LOCKED = TRUE` from its OPTIONS** — no rebuild needed,
and nothing is lost since the provider is already pinned by the BYOK gate.
Worth checking any other deployment setting that flag.


ESCRI11 (approved):
lgtm

